PR #5, opened 2025-10-27
Apply spotless plugin

adityamparikh:
Apply spotless plugin for consistent code formatting


adityamparikh:
@chatman @epugh what process do we follow for subsequent PR reviews?

PR #13, opened 2025-11-01
refactor: Fix jib build and docs

adityamparikh:
Update jib version.
Remove duplicate information from readme.
Remove tar building instructions.
Use BuildInfoReader instead of hardcoding jar and docker image name


epugh:
How do we get rid of the SonarQube check?

adityamparikh:
> How do we get rid of the SonarQube check?

Removed sonar check in https://github.com/adityamparikh/solr-mcp/pull/13

PR #17, opened 2025-11-05
feat: Add security and docs 

adityamparikh:
address https://github.com/apache/solr-mcp/issues/11. If merged, can incorporate into https://solr.apache.org/guide/solr/latest/


epugh:
I think this should actually be part of the Solr mono repo versus here...   we should be thinking about Solr 10.1 I suspect.   We still need to get release tooling done.    One idea...   we can publish this tutorial as a blog on Solr site whenever we are ready.   So maybe long term it's in the ref guide but short term a copy in the blog?

janhoy (commented):
Agree that the tutorial perhaps belongs in ref-guide once we have a stable release here. But for now I think it is ok to have it in this repo.

However, it should not be HTML format, perhaps Markdown or Asciidoc for easy move to ref-guide later? I believe GitHub renders both MD and AsciiDoc well when browsed on the website.

janhoy:
HTML is not friendly for future editing. I'd choose asciidoc as format and let GitHub render it for us for now. Then, moving it into the ref-guide (which is asciidoc) later would be easy.

PR #18, opened 2025-11-09
ci: add Testcontainers Docker checks and fix ATR release workflow

adityamparikh:
## Summary

- Add `@Testcontainers(disabledWithoutDocker = true)` to integration test classes so tests skip gracefully when Docker is unavailable
- Fix `atr-release-test.yml`: add `ASF_USERNAME` secret existence checks, update artifact upload action version

## Changed files

- `.github/workflows/atr-release-test.yml` — conditional checks for ASF secrets, artifact upload version fix
- 6 test classes — add `@Testcontainers(disabledWithoutDocker = true)` annotation

## Test plan

- [x] Tests skip cleanly when Docker daemon is not running
- [x] Tests run normally when Docker is available


.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,374 @@
...
+# - Optional inputs help document the vote thread and enable experimental signing
+on:
+  workflow_dispatch:
+    inputs:
+        # Semantic version of the approved release (no -rc suffix)
+      release_version:
+        description: 'Release version (e.g., 1.0.0)'
+        required: true
+        type: string
+        # Which release candidate was approved (used to check out the exact tag)
```
janhoy: The indentation of the comments are confusing, should be in line with the block below

janhoy (commented):
AI code tools are great, but I feel better discussing with a human than a raw AI output, there is always some level of noise in those reponses.

I have not reviewed the entire PR. So I just add some preliminary comments here. Some nice GH Action use for the release process, I like that.

epugh:
I don't have a lot of expertise in Docker builds.   It looks like the model in `solr-docker` is that the receipe, the Dockerfile, is checked in, it somehow goes to Dockerhub, and that the actual image is built in Docker hub?   Whereas, in apps I've done, I build the Docker image locally, and then push it up to Docker hub.   I *think* that is more the model that we have here, excpet instead of building the image locally, it would be done by GHA?

adityamparikh:
@epugh @janhoy Please see private@ for the public key and verify that everything works as expected.
from dfoulks[Drew Foulks](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=dfoulks)

Can I access it or one of you has to?

PR #19, opened 2025-11-14
deps: Spring AI 1.1.0 GA

adityamparikh:
Upgrade to Spring AI 1.1.0 GA
Spring Boot to 3.5.7

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,7 +28,7 @@ plugins {
 }
 
 group = "org.apache.solr"
-version = "0.0.1-SNAPSHOT"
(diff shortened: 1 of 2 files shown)


epugh (commented):
Can we separate the version uogrdrses from the other changes to simplify review?  Live tj see version bumps!

janhoy:
If two PRs truly depend on each other that’s ok, but here the upgrade is independent so better rebase on main with only the version change.

adityamparikh:
@epugh @janhoy Created https://github.com/apache/solr-mcp/pull/20/files

PR #21, opened 2025-11-16
feat: add oauth2 security for http mode

adityamparikh:
1. Following: https://github.com/spring-ai-community/mcp-security?tab=readme-ov-file#mcp-server-security, 
make the MCP Server an OAuth2 Resource Server and add mcp-security.
2. Disable Security related autoconfiguration for stdio mode.
3. As an example secure the search and indexing tools. Can be changed to finer grained authorizations such as solr:read, solr:write, solr:admin in the future. 
4. Added example run configurations to make it easier to run the application in Intellij Idea including one with OAUTH2_ISSUER_URI env variable.
5. Added Auth0 setup instructions. Could be switched with another issuer.
6. Added convenience scripts/get-auth0-token.sh script that will fetch the Bearer token from the authorization server (requires documented env variables set from 5). Added files that will contain env variable values and access token that will be obtained after running the script to .gitignore that should not be checked in.
7. Updated docs including screenshots of MCP inspector with authorization success and failure. Other MCP clients such as Claude desktop will follow.


janhoy:
I have been using this nice library in a couple of projects, including in [solr's `jwt-auth` module](https://github.com/apache/solr/blob/main/solr/modules/jwt-auth/src/test/org/apache/solr/security/jwt/JWTAuthPluginIntegrationTest.java#L115-L120): 

https://github.com/navikt/mock-oauth2-server

It is in fact written by a government agency in Norway and open-sourced :)

epugh:
I think in a merge we lost some code in this PR!

adityamparikh:
Closing this PR as the security code has already been merged to main through other PRs. Creating a new PR for the security bypass feature.

PR #23, opened 2025-12-13
feat: upgrade to Spring Boot 4.1.1 and Spring AI 2.0.1

adityamparikh:
Upgrades the server to **Spring Boot 4.1.1** and **Spring AI 2.0.1**. Both are GA on Maven Central; no milestone repositories are required.

> **Breaking configuration change (HTTP profile only):** OTLP export moves from gRPC on port 4317 to OTLP/HTTP on 4318, with one complete URL per signal. `OTEL_TRACES_URL` changes meaning — it was a *base* endpoint (`http://collector:4317`) and is now the *full* traces path (`http://collector:4318/v1/traces`). A value carried over unchanged does not error; traces simply stop arriving. `OTEL_METRICS_URL` and `OTEL_LOGS_URL` are new. See [docs/observability.md](docs/observability.md).

## Versions

| Component | `main` | This PR |
|---|---|---|
| Spring Boot | 3.5.14 | **4.1.1** |
| Spring AI | 1.1.7 | **2.0.1** |
| MCP Java SDK | 0.18.x (transitive) | **2.0.1**, pinned via `mcp-bom` (Spring AI 2.0.1 resolves 2.0.0; 2.0.1 fixes stateless-server error handling, which is our HTTP profile, and adds bounded reads on the HTTP and STDIO transports — drop the pin once Spring AI ships it) |
| `mcp-server-security` | 0.0.6 (built against Security 6.5 / `mcp-core` 0.x) | **0.1.14** (Security 7.1 / `mcp-core` 2.0) — catalog change only, `McpServerOAuth2Configurer` API unchanged |
| Testcontainers | 1.21.3 | **2.0.2** (module renames: `testcontainers-junit-jupiter`, `testcontainers-solr`, `testcontainers-grafana`) |
| CycloneDX Gradle plugin | 2.4.1 (pinned) | **3.4.1** (Boot 4.1.1's `CyclonedxPluginAction` recognises 3.x) |
| Apache RAT plugin | 0.8.1 | 0.8.2 |
| JSpecify | explicit dependency | dropped — built into Spring Boot 4 |

## Code changes

**Jackson 3.** `tools.jackson.databind` replaces `com.fasterxml.jackson.databind` in `SolrConfig`, `JsonResponseParser`, `JsonDocumentCreator`, `JsonUtils`, `CollectionService`, `SchemaService`. `JacksonException` replaces `IOException` / `JsonProcessingException` in catch clauses; `JsonNode.properties()` / `asString()` replace `fields()` / `asText()`. Annotations stay in `com.fasterxml.jackson.annotation`. The MCP SDK side uses `mcp-json-jackson3`, so the test clients build their transports with `io.modelcontextprotocol.json.jackson3.JacksonMcpJsonMapper` over `JsonMapper.builder().build()`.

**MCP annotations.** Package moved from `org.springaicommunity.mcp.annotation` to `org.springframework.ai.mcp.annotation` (now part of Spring AI core) in all four service classes. The MCP SDK 2.0 also validates request arguments at the boundary, so `CollectionServiceTest`'s null-completion case now asserts the SDK rejects a null `CompleteArgument` instead of exercising the service.

**Modular starters.** `spring-boot-starter-web` → `-webmvc` + `-json`; `-aop` → `-aspectj` (what `@Observed` needs on Boot 4); modular test starters (`-webmvc-test`, `-actuator-test`, `-opentelemetry-test`). `spring-ai-autoconfigure-mcp-server-common` and `-client-common` are declared explicitly: Spring AI 2.0.0-M7 marked them optional in the starter POMs, but the webmvc autoconfig still references types from them. `spring-boot-starter-mongodb` is excluded from Spring AI's docker-compose module, which otherwise tries to build a Mongo client on `bootRun`.

**Observability.** `spring-boot-starter-opentelemetry` (traces, metrics, OTLP log export) replaces the manual OTel instrumentation BOM 2.11.0 + `micrometer-tracing-bridge-otel` + Prometheus registry. The OTel logback appender (`2.21.0-alpha`) is installed programmatically by a new `InstallOpenTelemetryAppender` bean (`@Profile("http")`), and `application-http.properties` switches to `management.opentelemetry.*` / `management.otlp.*` endpoints. Two `resolutionStrategy` rules keep the OTel graph coherent: `opentelemetry-api-incubator` aligned to `1.62.0-alpha` (the appender pins `1.55.0-alpha`, which lacks `DeclarativeConfigProperties.get(String)` and fails at context startup), and `opentelemetry-proto` held at `1.3.2-alpha` (the `1.8.0-alpha` line needs protobuf 4). `@Observed` span names change from `search-service#search` to `SearchService#search` under the AspectJ starter; `DistributedTracingTest` is updated for the new names and disables export with the current `management.tracing.export.otlp.enabled` property. With the Prometheus registry gone, `/actuator/prometheus` is no longer exposed or documented; metrics leave via OTLP only.

**Profiles.** `application-stdio.properties` excludes the Boot 4.1 security auto-configuration classes (`org.springframework.boot.security.autoconfigure.SecurityAutoConfiguration` and `...actuate.web.servlet.ManagementWebSecurityAutoConfiguration`); the Boot 3 names would be silently ignored. `spring.application.name` stays `solr-mcp` in every profile.

**Logging under Spring Framework 7.** `application-stdio.properties` no longer sets `logging.pattern.console=`. Boot copies that value into the JVM-wide `CONSOLE_LOG_PATTERN` system property (first writer wins) and logback rejects an empty pattern, so it silenced nothing — but Framework 7 pauses a test's context on context switch and makes the next context re-initialise logback, so a stdio-profile test running first broke every later http-profile context in the same JVM. STDIO stays silent because `logback-spring.xml` declares no appenders for it (unchanged), and `LoggingConfigurationTest` gains a guard against the empty pattern coming back.

**GraalVM native.** `MemberCategory.DECLARED_FIELDS` (removal-marked) → `ACCESS_DECLARED_FIELDS`. New hints for the SolrJ 10 API model types (`SolrJerseyResponse`, `ErrorInfo`, `CoreStatusResponse` and nested), the relocated `org.springframework.ai.mcp.annotation.context.DefaultMetaProvider`, and Solr's `EnvToSyspropMappings.properties` / `DeprecatedSystemPropertyMappings.properties` resources. Error Prone / NullAway are disabled for the `compileAotJava` / `compileAotTestJava` tasks, whose generated sources cannot be edited.

**SBOM.** With cyclonedx 3.x the work splits into `cyclonedxDirectBom` (resolves, owns `includeConfigs`) and `cyclonedxBom` (aggregates). The direct task is scoped to `productionRuntimeClasspath`; left at defaults it adds ~100 test-only components that scanners would read as shipped. The license-notice plugin reads the SBOM from its new location, `build/reports/cyclonedx/application.cdx.json`. The old 2.4.1 pin turned out to ship **no** SBOM in the bootJar at all, because Boot's action bails on an unrecognised plugin version (#186).

**compose.yaml.** `grafana/otel-lgtm` pinned to `0.30.0`; Grafana and the OTLP receivers bind to loopback by default (`GRAFANA_BIND` / `OTLP_BIND` to override), anonymous access is Viewer rather than Admin (`GF_ANON_ROLE=Admin` to opt in); datasource list corrected to Prometheus, not Mimir. The service keeps its `org.springframework.boot.ignore` label, so Boot's compose support leaves it alone as on `main`.

**Docs.** `AGENTS.md` (framework versions, SBOM section, logging note, *Spring Boot 4 Notes* as migration notes), `docs/observability.md` (new endpoint scheme, per-signal variables, migration callout), new `dev-docs/Observability.md` (LGTM stack, architecture, production guidance), `dev-docs/graalvm-native-image.md` (OTel alignment replaces the old "BOM bump blocked" note), `docs/security/stdio.md`, `README.md`.

## Verification

| Command | Result |
|---|---|
| `./gradlew build` (Java 25) | 404 tests, 0 failures, 7 skipped |
| `./gradlew nativeTest -Pnative` (GraalVM CE 25.0.2) | 257 successful, 0 failed, 144 skipped (`main`: 227 / 0 / 142 — the delta is added tests, not skips) |
| `./gradlew dockerIntegrationTest` (JVM or native) | not run |

**Coverage caveat, unchanged from `main`:** no automated test exercises `McpServerOAuth2Configurer` on any version. It is wired only when `OAUTH2_ISSUER_URI` is set, and `DockerImageHttpIntegrationTest` runs without an issuer. #187 adds the first security-enabled context tests and applies here unchanged.

## References

- [Spring Boot 4.1 Release Notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.1-Release-Notes)
- [Spring AI 2.0.0 GA announcement](https://spring.io/blog/2026/06/12/spring-ai-2-0-0-GA-available-now)


adityamparikh:
Re-triggering CI now that ci.yml has landed on main (#117).

epugh:
Looks like you have some conflicts..    I think you need to update from main to pick up the new ci stuff?

adityamparikh:
Rebased onto `main` (`a84033b`) and now **green: 372 tests, 0 failures, 7 skipped** on Spring Boot 4.1.0 + Spring AI 2.0.0 GA.

**I have not force-pushed this PR's branch.** Its head is `adityamparikh:sb4`, which is the long-lived integration branch — 9 local branches track it (`chore/bump-*-sb4`, `feat/enable-virtual-threads-sb4`, `sb4-extras`, …) and rewriting it would strand all of them. The rebased result is on `adityamparikh:sb4-rebased-20260819` instead; say the word and I'll either retarget this PR's head to it or fast-forward `sb4` deliberately.

Four conflicts, all from the Spring AI 2.0 relocation meeting upstream's jspecify work:

- `CollectionService`/`SearchService`: took the `org.springframework.ai.mcp.annotation.*` imports **and** kept `org.jspecify.annotations.Nullable` — the merged bodies use `@Nullable` throughout.
- `SolrConfig`: dropped the Jackson 2 `ObjectMapper` import (Jackson 3's `tools.jackson.databind` supersedes it) but kept `java.net.URI`, which #176's path-normalisation fix needs.
- `build.gradle.kts`: kept **both** Error Prone blocks — upstream's `compileTestJava` NullAway disable and this branch's AOT-sources disable are independent.
- `SolrNativeHints`: kept both the SolrJ `EnvUtils` resource patterns and the `DefaultMetaProvider` registrations.

**Two things the rebase broke semantically** — both merged cleanly and failed only under javac/JUnit, so they are worth naming:

1. This branch had expanded the wildcard `Assertions.*` static import into an explicit list; upstream then added tests using `assertThrows` (#166) and `assertFalse`, which were not in that list. Compile failure.
2. `SolrNativeHintsTest` (new upstream) asserts the `DefaultMetaProvider` hint under its **pre-2.x** name `org.springaicommunity.mcp.context.DefaultMetaProvider`. Spring AI 2.x relocated it to `org.springframework.ai.mcp.annotation.context`. Because the registration uses `registerTypeIfPresent` — a no-op when the class is absent — the old name passes *vacuously* on 1.x and fails outright here. The assertion now names the class actually on this classpath. Worth flagging to reviewers: that test is weaker than it looks on `main` today.

Still parked behind the 1.0 release per the earlier discussion; this is just to keep it merge-ready.

adityamparikh:
Updated: `sb4` now points at the rebased, green state (`e01c7da`) — this PR shows it directly, so the temporary `sb4-rebased-20260819` branch is redundant.

The previous `sb4` tip (`cb0d94e`) is preserved on `adityamparikh:backup-sb4-pre-rebase-20260819` if anything needs recovering. Branches that were tracking the old `sb4` (`chore/bump-*-sb4`, `feat/enable-virtual-threads-sb4`, `sb4-extras`, …) will need rebasing onto the new tip.

adityamparikh:
Pushed 6364bd3 addressing review findings.

## Versions bumped

**4.1.0 → 4.1.1** and **2.0.0 → 2.0.1**, both checked against `maven-metadata.xml` directly (the
Maven Central search index returns stale, unsorted results for these coordinates). The PR
description called 4.1.0/2.0.0 "the latest GA releases" — true when written, no longer.
Description updated to match.

## The substantive finding: observability docs would have broken users

This PR moves OTLP export from **gRPC on 4317** to **HTTP on 4318**, with a separate full URL per
signal. That silently redefines `OTEL_TRACES_URL` — it used to be a *base* endpoint, and is now a
*complete* path ending in `/v1/traces`.

`docs/observability.md` — the user-facing guide README links to — was untouched by this PR and
still documented the removed `otel.exporter.otlp.*` properties plus:

```bash
export OTEL_TRACES_URL=https://otel-collector.example.com:4317
```

Following that after this merged would **not error**. Traces would just stop arriving, which is
the worst way for a telemetry config to fail.

Fixed:
- `docs/observability.md` — real property names, corrected production example, a variable table,
  and an explicit callout that the semantics changed for anyone upgrading
- `dev-docs/Observability.md` — this PR *adds* it, so it now states its relationship to the
  user-facing guide rather than silently competing with it; the three OTLP defaults are filled in
  instead of "(auto-configured)"
- `README.md` — names all four OTEL variables
- `AGENTS.md` — version references, plus the OTEL variables in Key Configuration

Also worth flagging as **breaking** in the release notes, not just the docs: anyone with
`OTEL_TRACES_URL` already set in a deployment is affected.

## Smaller items

- **Broken image removed.** `dev-docs/Observability.md` referenced `images/grafana-traces.png`,
  which does not exist anywhere in the repo — a broken-image icon on GitHub, introduced by this
  PR. The prose following it already describes the trace breakdown span by span, so it stands
  alone. Caught by a link check, not by reading.
- **The mongodb exclude is now explained.** It sat uncommented in a file where the cyclonedx pin
  above it gets fourteen lines of rationale — exactly the kind of thing someone deletes while
  tidying, and being `developmentOnly` it would break local `bootRun` without CI noticing.
- **CycloneDX pin now has a tracker** (#186) linked from both `build.gradle.kts` and `AGENTS.md`,
  with a checklist for verifying the SBOM still lands where `generateBinaryLicense` expects it.

## Verification

- `./gradlew build` — green, **372 tests / 0 failures**, matching `main`'s count exactly
- `./gradlew nativeTest -Pnative` — green, **227 passing / 0 failed**, unchanged by the bump

## Merge-order note

This PR and #175 both rewrite the same line in `JsonResponseParser.toNamedList` — this one changes
`fields().forEachRemaining(...)` to `properties().forEach(...)` for Jackson 3, #175 threads a node
path through it for the empty-facet fix. They compose (the `properties()` call *plus* the path
argument), but it is a guaranteed conflict. #175 is small and ready, so merging it first makes
this the cheaper rebase.

adityamparikh:
Verified the current head `d6e973d` (which merges `main` @ `b4ffe18`, including #193's logging rework) in a clean worktree:

| Command | Result |
|---|---|
| `./gradlew build` (JDK 25) | 404 tests, 0 failures, 7 skipped |
| `./gradlew nativeTest -Pnative` (GraalVM CE 25.0.2) | **257 successful, 0 failed, 144 skipped**, BUILD SUCCESSFUL in 8m 41s |

For comparison `main` measures 227 / 0 / 142 natively, so the delta is added tests, not skips. No conflicts with `main`; GitHub reports the PR mergeable.

PR #36, opened 2026-01-07
remove unneeded files

epugh:
Some files that have crept in that don't need to be in source control.


epugh:
I'll have to look...  sometimes if a person comments on a PR, then after that they can be added as an assignee or reviewer..

PR #99, opened 2026-04-24
fix: add Jackson annotations and defensive copying to SearchResponse

adityamparikh:
## Summary
- Add `@JsonInclude(NON_NULL)` and `@JsonIgnoreProperties(ignoreUnknown = true)` to `SearchResponse` for consistency with collection DTOs in `Dtos.java`
- Add compact constructor with defensive copying using `List.copyOf()` and `Map.copyOf()` for mutable `documents` and `facets` fields, defaulting to empty collections when null

## Test plan
- [x] No regressions

### DIFF
--- a/src/main/java/org/apache/solr/mcp/server/search/SearchResponse.java
+++ b/src/main/java/org/apache/solr/mcp/server/search/SearchResponse.java
@@ -16,6 +16,8 @@
  */
 package org.apache.solr.mcp.server.search;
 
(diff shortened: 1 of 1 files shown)


adityamparikh:
Closing: `SearchResponse` is `@NullMarked`, so the `documents != null ? … : List.of()` fallbacks guard against a null the type contract already excludes, and `@JsonIgnoreProperties(ignoreUnknown)` has no effect on a record that is only ever serialized. Not worth a PR.

PR #100, opened 2026-04-24
fix: use ObjectMapper for error JSON in SchemaService

adityamparikh:
## Summary
- Replace string interpolation of `e.getMessage()` in `getSchemaResource()` with proper ObjectMapper serialization to prevent JSON injection from exception messages containing special characters

## Test plan
- [x] No regressions


adityamparikh:
Closing as obsolete — this landed upstream via #176.

Current \`main\` (\`a84033b\`), `SchemaService.java:191`:

```java
return toJson(objectMapper, Map.of("error", String.valueOf(e.getMessage())));
```

That is this PR's change, and strictly better than it. This PR proposed:

```java
return toJson(objectMapper, Map.of("error", e.getMessage()));
```

`Map.of` rejects null values with an NPE, so on an exception whose `getMessage()` is null (common for NPE and several SolrJ wrappers) this PR's version would throw from inside the catch block and replace a JSON error payload with a propagated NPE. Upstream's `String.valueOf(...)` avoids exactly that.

No action needed — closing.

PR #103, opened 2026-04-24
refactor: reduce CollectionService internal method visibility

adityamparikh:
## Summary
- `buildIndexStats()`, `buildQueryStats()`, `getCacheMetrics()`, `getHandlerMetrics()` go from public to package-private. They are internal to `CollectionService`; every caller lives in `org.apache.solr.mcp.server.collection`, production and test alike (`CollectionServiceTest`, `CollectionServiceIntegrationTest`), so the narrowing costs no test access. `getCacheMetrics`/`getHandlerMetrics` keep their `@Nullable`.

## Verification
- `./gradlew build` and `./gradlew nativeTest -Pnative` pass; no test changes needed.

### DIFF
--- a/src/main/java/org/apache/solr/mcp/server/collection/CollectionService.java
+++ b/src/main/java/org/apache/solr/mcp/server/collection/CollectionService.java
@@ -464,7 +464,7 @@ public class CollectionService {
 	 * @see IndexStats
 	 * @see LukeResponse
 	 */
-	public IndexStats buildIndexStats(LukeResponse lukeResponse) {
+	IndexStats buildIndexStats(LukeResponse lukeResponse) {
 		NamedList<Object> indexInfo = lukeResponse.getIndexInfo();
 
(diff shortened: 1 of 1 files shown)


adityamparikh:
Rebased onto `main` (`a84033b`).

Conflict resolution: `main` had since marked `getCacheMetrics`/`getHandlerMetrics` as `public @Nullable` (they return `null` when the metrics endpoint is unavailable, e.g. Solr 10). I kept the `@Nullable` and applied only the visibility reduction, so both are now package-private *and* still annotated.

I also re-checked that the narrowing is actually safe: every caller of all four methods — `buildIndexStats`, `buildQueryStats`, `getCacheMetrics`, `getHandlerMetrics` — lives in `org.apache.solr.mcp.server.collection`, production and test alike (`CollectionServiceTest`, `CollectionServiceIntegrationTest`). Nothing outside the package refers to them, so package-private costs no test access.

PR #104, opened 2026-04-24
fix(config): reject a solr.url SolrJ cannot connect to at bind time

adityamparikh:
## Problem

`solr.url` is bound into a record and handed to SolrJ. Nothing checks that it is a URL SolrJ can use, so a misconfigured deployment starts cleanly and fails at first request with an opaque client error. The easy mistake is omitting the scheme:

```
solr.url=localhost:8983
  → URI.create(...) parses scheme="localhost", host=null
  → SolrConfig normalizes the *path* by string concatenation, never noticing
  → SolrJ: "localhost:8983/solr/" is not a URL it can connect to
```

## Change

`SolrConfigurationProperties` gets a compact constructor that rejects anything but an **absolute `http`/`https` URL with a host**, with a message that names the property, shows an example and echoes the offending value:

| Accepted | Rejected |
|---|---|
| `http://localhost:8983` | `localhost:8983` (scheme `localhost`, no host) |
| `http://localhost:8983/solr/` | `solr.example.com`, `/solr` (not absolute) |
| `https://solr.internal:8983/custom/solr/` | `ftp://…`, `file://…` (SolrJ cannot speak them) |
| `http://solr:8983/solr/` | `not a url`, `http://`, empty, blank, absent |

Boot surfaces the `IllegalArgumentException` through its bind failure analyzer, so startup prints *Failed to bind properties under 'solr'* with the reason.

**Why the constructor and not Bean Validation.** Hibernate Validator is not on the classpath; adding it for one field means a new runtime dependency in the native image plus two classes for a custom constraint, and `@NotBlank`/`@URL` alone cannot express "http(s) with a host" (Hibernate's `@URL` accepts `ftp://` and `file://`). A record's compact constructor is the plain-Java place for an invariant, and the binder invokes it reflectively with `null` when the property is absent, so the one `hasText` check is the guard for that boundary. No network I/O; native-safe as-is.

Path normalization (appending `/solr/`) is unchanged and still lives in `SolrConfig`.

## Verification

`SolrConfigurationPropertiesTest`: 19 cases over accepted URLs, rejected values, the absent value, and two `ApplicationContextRunner` bind tests. Those two are `@DisabledInNativeImage` because that runner builds its context through a JDK dynamic proxy; the constructor itself runs natively. `./gradlew build` on JDK 25 against `main` @ `b4ffe18`: green.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -121,6 +121,7 @@ dependencies {
     implementation(libs.spring.boot.starter.web)
     implementation(libs.spring.boot.starter.actuator)
     implementation(libs.spring.boot.starter.aop)
+    implementation(libs.spring.boot.starter.validation)
     implementation(libs.spring.ai.starter.mcp.server.webmvc)
     implementation(libs.solr.solrj)
     implementation(libs.commons.csv)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -148,6 +148,7 @@ dependencies {
     implementation(libs.spring.boot.starter.web)
     implementation(libs.spring.boot.starter.actuator)
     implementation(libs.spring.boot.starter.aop)
+    implementation(libs.spring.boot.starter.validation)
     implementation(libs.spring.ai.starter.mcp.server.webmvc)
     implementation(libs.solr.solrj)
     implementation(libs.commons.csv)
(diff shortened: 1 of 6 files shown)


adityamparikh:
Heads-up: this branch carries `777168e8` ("remove unused logback.xml"), which now conflicts with `main` — #189 made that file load-bearing and added a test that fails without it. #193 removes `logback.xml` properly (status listener set from `main()`, no `logging.config` override), so on the next rebase please drop `777168e8` and let #193 own the logging change. The URL-validation part of this PR is unaffected.

epugh (approved):

PR #108, opened 2026-04-24
fix: validate collection name consistently across all MCP tool methods

adityamparikh:
## Summary

Only `createCollection` and the two schema-modification tools validated the collection name. `checkHealth`, `getCollectionStats`, `search`, the four `index*Documents` tools, `getSchema` and `getSchemaResource` accepted null or blank and failed downstream, and a null collection reaching SolrJ silently targets the client's *default* collection rather than reporting a bad argument. The message was also duplicated: a constant in `CollectionService` and a copied literal in `SchemaService`.

## Changes

- One shared `ToolArguments.requireCollection(String)` in the `util` package, called by all twelve entry points, so the four services cannot drift apart. `SchemaService`'s private duplicate is removed.
- Single message: `"Collection name cannot be null or empty"`.

| Service | Methods validated |
|---|---|
| `CollectionService` | `createCollection`*, `getCollectionStats`, `checkHealth` |
| `IndexingService` | `indexJsonDocuments`, `indexCsvDocuments`, `indexXmlDocuments`, `indexMarkdownDocuments` |
| `SchemaService` | `getSchema`, `getSchemaResource`, `addFields`*, `addFieldTypes`* |
| `SearchService` | `search` |

\* already validated on `main`; switched to the shared helper.

## Why a runtime null check in `@NullMarked` code

NullAway is a compile-time analysis and only binds callers the compiler can see. MCP tool methods are invoked reflectively by the Spring AI annotation runtime, which looks each parameter up in the request's argument map and passes `null` through for a missing value. `@McpToolParam(required = true)` only marks the parameter required in the advertised JSON schema; the MCP SDK on `main` (0.18) does not validate incoming arguments against it. The rationale is recorded on `ToolArguments`.

`CollectionNameValidationTest` (10 tests) asserts every entry point against the shared constant rather than a copied literal.

Touches the same `SchemaService.getSchema` lines as #98; whichever merges second needs a one-line rebase.

### DIFF
--- a/src/main/java/org/apache/solr/mcp/server/collection/CollectionService.java
+++ b/src/main/java/org/apache/solr/mcp/server/collection/CollectionService.java
@@ -411,6 +411,10 @@ public class CollectionService {
 	public SolrMetrics getCollectionStats(
 			@McpToolParam(description = "Solr collection to get stats/metrics for") String collection)
 			throws SolrServerException, IOException {
+		if (collection == null || collection.isBlank()) {
+			throw new IllegalArgumentException(BLANK_COLLECTION_NAME_ERROR);
+		}
+
 		// Extract actual collection name from shard name if needed
 		String actualCollection = extractCollectionName(collection);
 
@@ -957,6 +961,10 @@ public class CollectionService {
 	 */
 	@McpTool(name = "check-health", description = "Check health of a Solr collection")
 	public SolrHealthStatus checkHealth(@McpToolParam(description = "Solr collection") String collection) {
+		if (collection == null || collection.isBlank()) {
+			throw new IllegalArgumentException(BLANK_COLLECTION_NAME_ERROR);
+		}
+
 		String actualCollection = extractCollectionName(collection);
 		try {
 			// Ping Solr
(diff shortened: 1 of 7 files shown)


adityamparikh:
Closing this. Re-checked against `main` @ `7d920af`, and most of the rationale in the description doesn't hold up:

- **"a null collection silently targets the client's *default* collection"** — not true here. `SolrConfig.buildSolrClient` never calls `withDefaultCollection`, so there is no default to fall back to.
- **whitespace needs a guard** — it doesn't. `" "` encodes to `%20`, reaches Solr and 404s, and `SearchService.withRemediationHint` already turns that into the `list-collections` hint. Same for any unknown collection name.

What's actually left is much narrower: only the *empty* string degrades. `/solr//select` is rejected by Jetty during URI parsing, so it is a **400 `Ambiguous URI empty segment`** rather than a 404 — it misses the `NOT_FOUND` branch, carries none of the message tokens the other branches match on, and reaches the client raw.

| collection argument | HTTP | what the client sees |
|---|---|---|
| `""` | 400 | `Ambiguous URI empty segment` |
| `" "` | 404 | + `list-collections` hint |
| unknown name | 404 | + `list-collections` hint |

That is a real dead end, but a rare one, and not worth the shape this PR has taken: eleven call sites plus ~140 lines of Mockito tests that only exercise the shared helper through four service constructors, and add ten `@DisabledInNativeImage` skips to `nativeTest`.

If it proves worth fixing later, the cheaper place is `SolrClient.request(SolrRequest<?>, String)` — the single abstract funnel every operation goes through, and where `null` legitimately means "no collection targeted" while blank never is. One check there would cover every existing and future tool.

The one piece worth keeping is the de-duplication: the same message exists twice, as `CollectionService.BLANK_COLLECTION_NAME_ERROR` and as a private `SchemaService.requireCollection`. I'll fold that into the next PR that touches `SchemaService`.

PR #111, opened 2026-04-24
fix: narrow catch clauses in fetchCacheMetrics/fetchHandlerMetrics

adityamparikh:
## Summary
- Narrow catch from `RuntimeException` to `SolrException` in `fetchCacheMetrics()` and `fetchHandlerMetrics()`
- Still catches Solr 10 `RemoteSolrException` (subclass of `SolrException`) for graceful degradation when `/admin/mbeans` is unavailable
- Programming bugs (NPE, ClassCastException, etc.) now propagate instead of being silently swallowed

## Test plan
- [x] No regressions

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -339,9 +339,11 @@ The Solr Docker image used in tests is configurable via the `solr.test.image` sy
 ### Solr 10 Compatibility
 
 Solr 10.0.0 is fully supported with the JSON wire format. The `/admin/mbeans` endpoint was
-removed in Solr 10; `getCacheMetrics()` and `getHandlerMetrics()` now catch `RuntimeException`
-(which covers `RemoteSolrException`) so they degrade gracefully and return `null`. Tests that
-check `cacheStats` and `handlerStats` already handle `null` values.
+removed in Solr 10; `getCacheMetrics()` and `getHandlerMetrics()` catch `SolrException` (which
+`RemoteSolrException` extends) so they degrade gracefully and return `null`. Tests that check
+`cacheStats` and `handlerStats` already handle `null` values. The catch is deliberately *not*
+`RuntimeException`: an unrelated runtime failure inside metrics parsing should surface as a bug,
+not be silently reported as "metrics unavailable".
 
 Remaining known differences from Solr 9:
 - **`/admin/mbeans` removed:** Cache and handler stats from `getCollectionStats()` will always be `null` on Solr 10. A future migration to `/admin/metrics` will restore these metrics.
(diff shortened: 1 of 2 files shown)


adityamparikh:
Rebased onto `main` (`a84033b`), with one fix to the PR's own content.

**Fixed:** the diff changed `catch (... | RuntimeException _)` to `catch (... | SolrException e)` — naming a variable that is never used. `main` deliberately uses the unnamed `_` there. Restored to `catch (SolrServerException | IOException | SolrException _)` so the narrowing lands without reintroducing an unused binding.

**One thing worth a reviewer's judgement.** `RemoteSolrException extends SolrException extends RuntimeException`, so the 404 from a removed endpoint is still caught and still degrades to `null` — that part is safe. But `fetchMetrics` also does an unchecked `(NamedList<Object>) response.get("metrics")`. Under the old broad `RuntimeException` catch, a `ClassCastException` from an unexpected `/admin/metrics` response shape would degrade to `null`; after this narrowing it propagates out of `getCollectionStats`.

I think that is the right trade — swallowing `ClassCastException` hides a genuine bug, and surfacing it is the stated point of the PR — but it is a real behaviour change on any Solr whose metrics payload differs from 9.9, which is what our default `solr.test.image` exercises. Happy to add `ClassCastException` back to the catch list if the project would rather keep unconditional graceful degradation there.

adityamparikh:
Two updates.

**Added the doc correction this PR invalidates (ab1aa24).** The Solr 10 Compatibility section said
`getCacheMetrics()`/`getHandlerMetrics()` "catch `RuntimeException` (which covers
`RemoteSolrException`)" — untrue the moment this merges, so the fix belongs here rather than
trailing behind. It now also records *why* the narrower catch is correct, which the old wording
didn't. (`CLAUDE.md` is a symlink to `AGENTS.md`, so that's the single source.)

**Resolved a collision with #110.** Both PRs rewrite the same two catch clauses — this one narrows
`RuntimeException` to `SolrException`, #110 was binding the exception for logging while leaving
`RuntimeException` in place. Whichever merged second would have conflicted or silently reverted
the other. #110 now carries the narrowed form too (506f2bb), so the end state is identical in
either merge order and neither blocks the other.

For the record, the narrowing is safe: `RemoteSolrException extends SolrException` in solrj
10.0.0, verified against the jar rather than from memory, so Solr 10's missing `/admin/mbeans`
still degrades to `null`. What stops being swallowed is unrelated runtime failures — which is the
point.

adityamparikh:
Closing: superseded by #110, which merged the same `SolrException` narrowing in both catch clauses. The only remaining piece is the AGENTS.md sentence that still says `RuntimeException`; that one-liner will go in a separate docs fix.

PR #115, opened 2026-04-24
fix: add JSpecify @Nullable annotations for accurate null contracts

adityamparikh:
## Summary
- Add `@Nullable` annotations from JSpecify to all locations where null is a legitimate value
- Fixes the mismatch between the `@NullMarked` package declaration and actual null usage
- Enables NullAway to accurately detect null-safety violations
- Covers DTOs, service return types, and utility method signatures

## Test plan
- [x] `./gradlew build` passes (including NullAway checks)
- [x] No regressions


epugh:
@adityamparikh I updated this PR for the new `ci` integration tests...    We may want to think about if our integration tests are reliable to run as part of ci.    Can you assess if the failure is related to this PR or a flaky test and something we need to harden.   If tests are flaky, we probably don't want to add then to CI, they may offer more frustration than value!

adityamparikh:
Closing as superseded by #133, which is a strict superset for the JSpecify rollout: all `Dtos.java`, `CollectionUtils.get{Long,Float,Integer}`, `CollectionService` cache/handler return types, and `JsonResponseParser.convertValue` annotations from this PR are included there, along with @NullMarked package-info for every sub-package.

Note: the `SolrConfig#useHttp1_1(true)` change in this PR is an unrelated HTTP wire-protocol fix and is **not** in #133. If we still want it, it should land as its own focused PR rather than ride along with the JSpecify rollout.

adityamparikh:
Correction to my previous note: the `useHttp1_1(true)` change landed separately via #130 (commit 62e92cd) and is already on `main`. So this PR is fully subsumed by #133 — nothing left to extract.

PR #122, opened 2026-05-02
fix(security): prevent Solr local-param injection in SearchService.search

adityamparikh:
## Summary

\`SearchService.search\` previously passed the user's \`query\` argument straight to Solr's standard query parser via \`solrQuery.setQuery(query)\`. Solr's standard parser honors \`{!parser arg=val}\` local-param prefixes, which lets a caller switch parsers mid-query — including:

- \`{!xmlparser ...}\` — historical XXE/RCE ([CVE-2017-12629](https://nvd.nist.gov/vuln/detail/CVE-2017-12629))
- \`{!join from=… fromIndex=other …}\` — cross-collection reads
- \`{!func}\` / \`_val_:\` — DoS-grade function plans

This PR binds the user's input to a separate \`qq\` request parameter and references it from a constant \`q={!edismax v=\$qq}\`, so the standard parser never sees raw user input. eDisMax does not honor a \`{!parser …}\` prefix inside its input, so the injection vectors are treated as literal characters.

Reference: [Solr Reference Guide — Parameter Substitution](https://solr.apache.org/guide/solr/latest/query-guide/common-query-parameters.html#parameter-substitution).

## Test plan
- [x] Five new unit tests in \`SearchServiceTest\` cover blank query, plain query, \`{!xmlparser}\` injection, \`{!join}\` injection, and \`_val_:\` function-query injection — all assert the malicious string lands in \`qq\`, not in \`q\`
- [x] Three existing tests updated to match the new q/qq construction
- [x] \`./gradlew test --tests \"*SearchService*\"\` passes
- [x] \`./gradlew build\` passes (full test suite)

## References
- [CWE-943: Improper Neutralization of Special Elements in Data Query Logic](https://cwe.mitre.org/data/definitions/943.html)

### DIFF
--- a/src/main/java/org/apache/solr/mcp/server/search/SearchService.java
+++ b/src/main/java/org/apache/solr/mcp/server/search/SearchService.java
@@ -196,8 +196,14 @@ public class SearchService {
 	 * @param collection
 	 *            The Solr collection to query
 	 * @param query
-	 *            The Solr query string (q parameter). Defaults to "*:*" if not
-	 *            specified
+	 *            The user-supplied search expression. To prevent local-param
+	 *            injection (e.g. {@code {!xmlparser ...}}, {@code {!join ...}},
+	 *            {@code {!func}}/{@code _val_}), this value is bound to the
+	 *            {@code qq} request parameter and dereferenced from a constant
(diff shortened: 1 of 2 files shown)


epugh:
i think these we don't want to merge in..  let's make Solr responsible for security.....

PR #128, opened 2026-05-04
ci: run JVM + native Docker integration tests on every PR

adityamparikh:
## Summary

- Adds a `JVM Docker Integration` job to `ci.yml` that runs both `DockerImageStdioIntegrationTest` and `DockerImageHttpIntegrationTest` against the Jib-built JVM image on every PR.
- Drops the `paths:` filter from `native.yml` so `nativeTest`, the native Docker integration job, and the JVM-vs-native benchmark run on every PR (not only PRs that touch native-related files).
- Removes the `DockerImageHttpIntegrationTest` exclusion for `-Pnative` builds and the `bootBuildImage` AOT profile lock (plus the `processAot` profile-lock block) so the native image actually serves HTTP at runtime.
- Lifts the `SecurityAutoConfiguration` / `ManagementWebSecurityAutoConfiguration` exclusions out of `application-stdio.properties` into a class-level `@SpringBootApplication(excludeName=...)` on `Main.java`, the prerequisite for unlocking AOT.
- Replaces a flawed log-grep assertion in `DockerImageHttpIntegrationTest.testHttpModeConfiguration` with a TCP-layer reachability check. The previous assertion checked for "Tomcat started on port" in `docker logs`, which never appeared because the http-profile logback config pipes Spring Boot's startup logs through a CONSOLE + OpenTelemetryAppender pair whose init state suppresses stdout when the OTLP collector is unreachable. The replacement opens a raw socket to the mapped port — a true positive proxy for "HTTP mode is active."

End state: every PR exercises four Docker integration combinations — JVM Jib + STDIO, JVM Jib + HTTP, native buildpack + STDIO, native buildpack + HTTP — plus the existing JVM unit and Testcontainers integration tests, `nativeTest`, and the JVM-vs-native benchmark.

## Test plan

- [x] `./gradlew clean build dockerIntegrationTest` passes locally on a Linux/macOS dev box. 11 docker tests = 4 STDIO + 7 HTTP, JVM Jib image.
- [x] `./gradlew bootBuildImage -Pnative && ./gradlew dockerIntegrationTest -Pnative` passes locally. Same 11 docker tests against the native buildpack image.
- [ ] CI: build, unit-tests, integration-tests, jvm-docker-tests, sonarcloud all green.
- [ ] CI: nativeTest, native-image, benchmark all green.

## Notes for reviewers

- The `OpenTelemetryAppender` install glue may be missing on this branch (no Spring bean appears to call `OpenTelemetryAppender.install(openTelemetry)` after the SDK initializes), which is why Spring Boot's INFO startup logs don't reach `docker logs` under the http profile. That's a pre-existing logging issue, not introduced here. Filing as a follow-up; this PR's scope is intentionally limited to CI plumbing + native HTTP enablement.
- The class-level `@SpringBootApplication(excludeName=...)` change is identical in shape to a change already living on the `sb4` Spring Boot 4 fork, which has been running with this configuration since May 1.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,38 @@ jobs:
                     retention-days: 7
 
     # ========================================================================
-    # Job 4: Solr Version Compatibility
+    # Job 4: JVM Docker Integration
+    # ========================================================================
+    # Runs DockerImageStdioIntegrationTest and DockerImageHttpIntegrationTest
+    # against the JVM Docker image produced by Jib. Validates that the
+    # production container starts and serves both transports correctly.
+    # ========================================================================
+    jvm-docker-tests:
+        name: JVM Docker Integration
+        runs-on: ubuntu-latest
+        needs: build
+        timeout-minutes: 30
+
+        steps:
+            -   name: Checkout code
+                uses: actions/checkout@v4
+
+            -   name: Set up Java
+                uses: ./.github/actions/setup-java
+
+            -   name: Run JVM Docker integration tests
+                run: ./gradlew dockerIntegrationTest --no-daemon
+
+            -   name: Upload test results
+                if: always()
+                uses: actions/upload-artifact@v4
+                with:
+                    name: jvm-docker-test-results
+                    path: build/reports/dockerIntegrationTest
+                    retention-days: 7
+
+    # ========================================================================
+    # Job 5: Solr Version Compatibility
     # ========================================================================
     # Tests against multiple Solr versions to catch compatibility regressions.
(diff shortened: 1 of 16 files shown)


adityamparikh:
Reframing this PR to target the in-flight Docker image matrix work (Jib JVM + per-profile Paketo native images). The new PR builds on top of that branch and updates the suffix scheme + javadoc accordingly:

→ https://github.com/adityamparikh/solr-mcp/pull/80 (against `adityamparikh:fix/native-test-integration-tests`)

Once the parent branch lands on `apache:main`, the new PR will be retargeted upstream as a fresh PR here.

PR #132, opened 2026-05-18
refactor(dtos): drop noise fields from MCP tool response records

adityamparikh:
## Summary

Three MCP tool response records in \`Dtos.java\` carried fields that conveyed no real information to the LLM client. Removing them simplifies the JSON wire format and eliminates legacy \`java.util.Date\` usage entirely.

## Fields removed

| Record | Field | Why noise |
|---|---|---|
| \`SolrMetrics\` | \`timestamp\` (Date) | Always \"now\" — metrics collected synchronously per call. MCP host records call timing already. |
| \`SolrHealthStatus\` | \`lastChecked\` (Date) | Same: always \"now\". |
| \`SolrHealthStatus\` | \`solrVersion\` (String) | Always null in every code path. \`@JsonInclude(NON_NULL)\` was hiding it on the wire anyway. |
| \`SolrHealthStatus\` | \`status\` (String) | Same: always null. |
| \`CollectionCreationResult\` | \`success\` (boolean) | Always \`true\` on return — failures throw before reaching the result. |
| \`CollectionCreationResult\` | \`message\` (String) | Always the literal \"Collection created successfully\". |
| \`CollectionCreationResult\` | \`createdAt\` (Date) | Always \"now\". |

After this change:
- \`SolrMetrics\` → \`(indexStats, queryStats, cacheStats, handlerStats)\`
- \`SolrHealthStatus\` → \`(isHealthy, errorMessage, responseTime, totalDocuments, collection)\`
- \`CollectionCreationResult\` → \`(name)\`

## Date → Instant migration: not needed

The original motivation was \"why are we using \`java.util.Date\` instead of \`java.time.Instant\`?\" — once every \`Date\` field turned out to be cargo-culted noise rather than load-bearing state, the migration is moot. Both \`java.util.Date\` and \`com.fasterxml.jackson.annotation.JsonFormat\` imports drop from \`Dtos.java\` entirely.

## Wire format impact

For LLM clients consuming the existing JSON, the change is purely subtractive — three keys disappear from each affected response. Any code that depended on those values (none in this repo) would break, but since \`@JsonInclude(NON_NULL)\` was already dropping \`solrVersion\` and \`status\` from the wire, only \`timestamp\` / \`lastChecked\` / \`success\` / \`message\` / \`createdAt\` are visible behavioral changes.

## Test plan

- [x] \`./gradlew clean build\` — BUILD SUCCESSFUL, all tests pass
- [x] \`./gradlew nativeTest -Pnative\` — 157/157 tests pass, 0 failures
- [x] Test assertions updated where they referenced removed fields (\`CollectionServiceIntegrationTest\`, \`CollectionServiceTest\`, \`ConferenceEndToEndIntegrationTest\`, \`McpClientIntegrationTestBase\`)
- [ ] CI native + docker matrix on the PR

## Related

This addresses a sub-thread of review on #131 (schema modification PR), where the new \`SchemaUpdateResult\` followed the same convention and the reviewer asked why we used \`Date\`. After this PR merges, #131 will pick up the same treatment for \`SchemaUpdateResult\` via rebase.

### DIFF
--- a/src/main/java/org/apache/solr/mcp/server/collection/CollectionService.java
+++ b/src/main/java/org/apache/solr/mcp/server/collection/CollectionService.java
@@ -26,7 +26,6 @@ import io.micrometer.observation.annotation.Observed;
 import io.modelcontextprotocol.spec.McpSchema.CompleteRequest;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import org.apache.solr.client.solrj.SolrClient;
@@ -525,7 +524,7 @@ public class CollectionService {
 		QueryResponse statsResponse = solrClient.query(actualCollection, new SolrQuery(ALL_DOCUMENTS_QUERY).setRows(0));
 
 		return new SolrMetrics(buildIndexStats(lukeResponse), buildQueryStats(statsResponse),
-				fetchCacheMetrics(actualCollection), fetchHandlerMetrics(actualCollection), new Date());
+				fetchCacheMetrics(actualCollection), fetchHandlerMetrics(actualCollection));
 	}
 
 	/**
@@ -1069,10 +1068,10 @@ public class CollectionService {
 					new SolrQuery(ALL_DOCUMENTS_QUERY).setRows(0));
 
 			return new SolrHealthStatus(true, null, pingResponse.getElapsedTime(),
-					statsResponse.getResults().getNumFound(), new Date(), actualCollection, null, null);
+					statsResponse.getResults().getNumFound(), actualCollection);
 
 		} catch (Exception e) {
-			return new SolrHealthStatus(false, e.getMessage(), null, null, new Date(), actualCollection, null, null);
+			return new SolrHealthStatus(false, e.getMessage(), null, null, actualCollection);
 		}
 	}
 
@@ -1136,7 +1135,7 @@ public class CollectionService {
 		CollectionAdminRequest.createCollection(name, effectiveConfigSet, effectiveShards, effectiveRf)
 				.process(solrClient);
 
-		return new CollectionCreationResult(name, true, "Collection created successfully", new Date());
+		return new CollectionCreationResult(name);
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


adityamparikh:
Closing this as superseded by #106 and #113, both of which have now landed on `main`.

This PR originated from a review sub-thread on #131 asking why the DTO records used
`java.util.Date` instead of `java.time.Instant`. My answer here was "the migration isn't
needed, because every `Date` field is cargo-culted noise that should just be deleted."
The project went the other way, and I think that's the better call:

- **#113** (`refactor: replace mutable Date with Instant in DTO records`) performed the
  migration directly, keeping the timestamps and modernizing their type. That resolves the
  original review question.
- **#106** (`refactor: remove dead fields and unused FieldStats type`) already removed
  `SolrHealthStatus.solrVersion` and `SolrHealthStatus.status` — two of the seven fields
  this PR targeted.

What remained unique to this PR was removing `CollectionCreationResult.success` / `.message`,
plus deleting `SolrMetrics.timestamp`, `SolrHealthStatus.lastChecked` and
`CollectionCreationResult.createdAt` — the three fields #113 had just deliberately converted
to `Instant`. Rebasing would mean resolving those conflicts by reverting a considered decision
that was merged on purpose, which isn't a merge conflict so much as a disagreement, and not
one worth relitigating for the wire-format savings involved.

If the `CollectionCreationResult.success` / `.message` removal is still wanted on its own
merits (both are constant — `success` is always `true` because failures throw before the
result is built, and `message` is always the literal `"Collection created successfully"`),
that's a much smaller, self-contained change and I'm happy to open a fresh PR scoped to just
that. Leaving it out for now rather than assuming.

One incidental finding worth recording: after rebasing this branch onto current `main`, the
build went red in `McpClientIntegrationTestBase.createShowsCollection()`. #131 added that test
after this branch was cut, and it reintroduced the same
`text.contains("success") || text.contains("true")` assertion pattern that this PR invalidates.
The rebase was textually clean because the two changes touch different lines of the same file.
That assertion is fine on `main` today, since `CollectionCreationResult.success` still exists —
but it's coupled to a field's presence rather than to the tool actually succeeding, so it would
break again under any future change that drops the field. Asserting on the echoed collection
name is the more durable form.

PR #143, opened 2026-06-11
docs(site): host MCP documentation content in-repo, deployed to solr-site at build time

adityamparikh:
## Summary

Makes **solr-mcp the source of truth for the MCP documentation content**, so docs travel with the code — every feature PR can update its docs in the same review. This addresses @janhoy's request on [apache/solr-site#175](https://github.com/apache/solr-site/pull/175#issuecomment-4664039372): *"Ideally I'd like the mcp docs to live in the same repo as code, but be deployed to solr site."*

Adds under `docs/site/content/`:
- `pages/mcp/**` — the MCP documentation Markdown pages (Quick Start, per-client setup, Security, Observability, Features, Resources, …)
- `doap/solr-mcp.rdf` — the DOAP project descriptor

Plus `docs/site/README.md` describing the content/presentation split and the build-time assembly contract.

## How it gets deployed

Content here is **plain Pelican Markdown** (front-matter selects a `template: mcp/*`). The **presentation layer** (Pelican templates, theme, CSS, `pelicanconf.py`) stays in [apache/solr-site](https://github.com/apache/solr-site), so the MCP pages keep the shared Solr site look and their published URLs (`https://solr.apache.org/mcp/...`) are unchanged.

At site-build time, solr-site fetches this directory into its Pelican `content/` tree before rendering — mirroring how the Solr Reference Guide content is assembled into the site. That wiring (build.sh + Pelican CI + removal of the now-sourced content) is the companion change in **apache/solr-site#175**.

## Sequencing

The solr-site fetch pulls from `apache/solr-mcp` `main`, so this PR should land first; then apache/solr-site#175 builds against it.

## Notes
- Design doc: `docs/superpowers/specs/2026-06-09-license-notice-and-docs-deploy-design.md`.
- No code or build changes — documentation content only.


janhoy:
I'm tempted to disable the feature that "intelligently" converts SOLR-NNN in text to JIRA links, as that feature is not intelligent at all.

janhoy (changes requested):
This branch contains also the LICENSE/NOTICE commit. Please remove.

### Important clarifications

* Are we not going to have versioned documentation with each release?
* What is "project" site and what is "product" docs? They seem conflated here now.

Let's continue on dev@ list these discussions

adityamparikh:
Thanks for the review @janhoy.

**LICENSE/NOTICE commit — done.** I force-pushed to drop the commit that modified the repo-root `LICENSE`/`NOTICE` files, so this PR no longer touches any root license file (`git diff main --name-only` confirms docs-content only). The only license-related material left is *content/design*, not the repo NOTICE: a `licensing.md` doc page and the design spec under `docs/superpowers/specs/`. Happy to split those out as well if you'd prefer this PR carry zero licensing material.

**Versioned docs per release** and **"project site" vs "product docs"** — agreed these are the larger design questions rather than something to settle in this diff; let's continue them on dev@ as you suggested.

While here I also tightened up accuracy in the MCP pages: corrected the `claude mcp add` syntax and per-client config, fixed broken doc links, added the Linux Docker note across clients, refreshed the supported Solr versions, and documented completions.

epugh:
@janhoy I hope I didn't get too merge crazy, and that we are all ont he same page about where docs live...   When solr-operator adopted the new ref guide based appraoch, I wondered if we should be following what they did: https://solr.apache.org/guide/operator/latest/index.html

I like that when I am in solr docs and search for something operator related it comes up!  
image

PR #145, opened 2026-06-12
ci: fix workflows blocked by the ASF GitHub Actions allow-list

adityamparikh:
## Problem

Every run of **Native Image** (`native.yml`) shows `startup_failure` and **build-and-publish.yml** shows `failure`, on every branch including `main` — with valid YAML and zero check-runs. The annotation on the run page is:

> The action `graalvm/setup-graalvm@v1` is not allowed in apache/solr-mcp because all actions must be from a repository owned by your enterprise, created by GitHub, or match one of the patterns…

## Root cause

The `apache` org [allow-lists Actions](https://infra.apache.org/github-actions-policy.html) and matches third-party actions by **exact commit SHA**. These workflows referenced third-party actions by **mutable tag** (`graalvm/setup-graalvm@v1`, `docker/login-action@v3`, `docker/setup-buildx-action@v3`, `softprops/action-gh-release@v1`, `peter-evans/repository-dispatch@v2`). A tag never matches a SHA-based allow-list entry, so the whole workflow fails to load before any job starts.

Notably `graalvm/setup-graalvm` **is already allow-listed** — at specific SHAs in [`apache/infrastructure-actions`](https://github.com/apache/infrastructure-actions/blob/main/actions.yml). The `@v1` tag just resolved to a commit that wasn't one of them. So **no infrastructure-actions PR is needed** — pinning to an already-listed SHA is sufficient.

(`CI` showing `action_required` is unrelated — that's the normal fork-PR approval gate.)

## Fix — SHA-pin the allow-listed actions, inline the rest

| Workflow | Change |
|---|---|
| `native.yml`, `release-publish.yml` | pin `graalvm/setup-graalvm` → `@329c42c…` (**v1.5.4, no `expires_at`** — chosen over v1.5.3 which expires 2026-08-28) |
| `build-and-publish.yml`, `release-publish.yml` | pin `docker/login-action` → `@650006c6…` (**v4.2.0, no `expires_at`** — already ASF-allow-listed) |
| `release-publish.yml` | drop `docker/setup-buildx-action` (default buildx handles `docker buildx imagetools create`) |
| `nightly-build.yml`, `atr-release.yml` | `softprops/action-gh-release` → `gh release create --notes-file` |
| `atr-release.yml` | `peter-evans/repository-dispatch` → `gh api repos/<r>/dispatches --input -` |

`graalvm/setup-graalvm` and `docker/login-action` are both on the ASF allow-list, so they're kept and pinned to a non-expiring SHA. `peter-evans/repository-dispatch` is **not** on the list and `docker/setup-buildx-action` is unnecessary for `imagetools create`, so those are inlined/dropped. `${{ }}` contexts moved onto `run:` lines are passed via `env:` and referenced as `"$VAR"` to avoid shell injection.

After this change the only non-`actions/*`/`apache/*` actions in the tree are the SHA-pinned, allow-listed `graalvm/setup-graalvm` and `docker/login-action`; everything else is inline `gh`/`docker` or the local `./.github/actions/setup-java` composite.

## Note

This unblocks `native.yml` from *starting*. The `nativeTest` job may then surface a separate, pre-existing native-test failure (it's been masked by `startup_failure` and also fails on `main`); the `native-image` + `benchmark` jobs and the JVM workflows should go green.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -274,11 +274,10 @@ jobs:
             # Authenticate to GitHub Container Registry
             # Uses built-in GITHUB_TOKEN (no configuration needed)
             -   name: Log in to GitHub Container Registry
-                uses: docker/login-action@v3
-                with:
-                    registry: ghcr.io
-                    username: ${{ github.actor }}
-                    password: ${{ secrets.GITHUB_TOKEN }}
+                env:
+                    REGISTRY_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                    REGISTRY_USER: ${{ github.actor }}
+                run: echo "$REGISTRY_TOKEN" | docker login ghcr.io -u "$REGISTRY_USER" --password-stdin
 
             # Authenticate to Docker Hub
             # Requires DOCKERHUB_USERNAME and DOCKERHUB_TOKEN secrets
@@ -307,10 +306,7 @@ jobs:
             #
             #      Note: `GITHUB_TOKEN` is provided automatically for GHCR; do not store it manually.
             #      - name: Log in to Docker Hub
-            #        uses: docker/login-action@v3
-            #        with:
-            #          username: ${{ secrets.DOCKERHUB_USERNAME }}
-            #          password: ${{ secrets.DOCKERHUB_TOKEN }}
+            #        run: echo "${{ secrets.DOCKERHUB_TOKEN }}" | docker login -u "${{ secrets.DOCKERHUB_USERNAME }}" --password-stdin
 
             # Convert repository owner to lowercase
             # Required because container registry names must be lowercase
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -274,10 +274,11 @@ jobs:
             # Authenticate to GitHub Container Registry
             # Uses built-in GITHUB_TOKEN (no configuration needed)
             -   name: Log in to GitHub Container Registry
-                env:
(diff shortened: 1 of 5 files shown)


.github/workflows/build-and-publish.yml
```diff
@@ -274,11 +274,10 @@ jobs:
             # Authenticate to GitHub Container Registry
             # Uses built-in GITHUB_TOKEN (no configuration needed)
             -   name: Log in to GitHub Container Registry
-                uses: docker/login-action@v3
```
janhoy, line 277: I see `docker-login-action` in the list of pre-approved actions, so can we instead use one of those hashes?

janhoy (commented):
Looks decent, one comment

.github/workflows/build-and-publish.yml
adityamparikh: Done in 7b1c810 — pinned `docker/login-action` to `650006c6eb7dba73a995cc03b0b2d7f5ca915bee` (v4.2.0), the entry with no `expires_at` in [`apache/infrastructure-actions/actions.yml`](https://github.com/apache/infrastructure-actions/blob/main/actions.yml), so it won't need re-bumping when older SHAs roll off. Applied to both GHCR logins (`build-and-publish.yml` + `release-publish.yml`) and the commented-out Docker Hub example.

For the others: `softprops/action-gh-release` (allow-listed via `'*': keep: true`) and `peter-evans/repository-dispatch` (not on the list at all) stayed as inline `gh` to keep the third-party surface minimal, and `docker/setup-buildx-action` stayed dropped since `docker buildx imagetools create` works with the default builder. Happy to restore `action-gh-release` to a pinned action too if you'd rather be consistent.

PR #149, opened 2026-06-13
Add RAT for header checking

### LINKED ISSUE #141: Add Apache RAT for license scanning
To enforce proper license headers, consider adding RAT. Gradle plugin: https://github.com/eskatos/creadur-rat-gradle

epugh:
Fixes #141

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,6 +28,7 @@ plugins {
     alias(libs.plugins.jib)
     alias(libs.plugins.graalvm.native) apply false
     alias(libs.plugins.cyclonedx)
+    alias(libs.plugins.rat)
 }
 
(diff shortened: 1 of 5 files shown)


adityamparikh:
Alternative approach for comparison: #150 implements this same RAT header enforcement as a `buildSrc` convention plugin (stacked on #138's buildSrc infra) instead of inline `tasks.rat { }` config. It also pulls the `.gitignore`→glob translation into a unit-tested `RatExcludes` helper, which caught an interior-slash anchoring bug in the inline version (`src/generated` was matched at any depth rather than root-anchored).

Whichever direction is preferred — inline (here) or buildSrc (#150) — happy to converge on one.

epugh:
close in favour of #150

PR #156, opened 2026-06-19
Prune out ai docs, but keep details on graalvm.

### LINKED ISSUE #148: Figure out what to do with AI coding docs
We have /docs/specs and docs/superpowers.   Some questions:

1. Do we even want to keep them?
2. Should we put the valuable evergreen content into /dev-docs?
3. WOuld we want to keep them, but put them under /dev-docs?

epugh:
This prunes the AI generated docs that are really just for the process.  The graalvm area is both in progress, and area that we need to share a lot of details, so adding to dev-docs/


adityamparikh:
LGTM!

PR #159, opened 2026-06-22
feat(collection): added alias management

y-luis-rojo:
feat(collection): add alias management tools (list, create, delete)

### DIFF
--- a/src/main/java/org/apache/solr/mcp/server/collection/AliasResult.java
+++ b/src/main/java/org/apache/solr/mcp/server/collection/AliasResult.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.mcp.server.collection;
+
+import java.util.Date;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Result record for alias management operations.
+ *
+ * <p>
+ * Returned by {@link AliasService} methods to communicate the outcome of
+ * create, update, and delete operations on Solr aliases.
+ *
+ * @param aliasName
+ *            the alias that was operated on
+ * @param collections
+ *            the target collection(s) (null for delete operations)
+ * @param success
+ *            whether the operation completed successfully
+ * @param message
+ *            human-readable description of the outcome
+ * @param timestamp
+ *            when the operation was performed
+ */
+public record AliasResult(String aliasName, @Nullable String collections, boolean success, String message,
+		Date timestamp) {
+}
--- a/src/main/java/org/apache/solr/mcp/server/collection/AliasResult.java
+++ b/src/main/java/org/apache/solr/mcp/server/collection/AliasResult.java
@@ -17,9 +17,7 @@
 package org.apache.solr.mcp.server.collection;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
-
 import java.util.Date;
-
 import org.jspecify.annotations.Nullable;
 
 /**
@@ -29,12 +27,17 @@ import org.jspecify.annotations.Nullable;
  * Returned by {@link AliasService} methods to communicate the outcome of
  * create, update, and delete operations on Solr aliases.
  *
- * @param aliasName   the alias that was operated on
- * @param collections the target collection(s) (null for delete operations)
- * @param success     whether the operation completed successfully
- * @param message     human-readable description of the outcome
- * @param timestamp   when the operation was performed
+ * @param aliasName
+ *            the alias that was operated on
+ * @param collections
+ *            the target collection(s) (null for delete operations)
+ * @param success
+ *            whether the operation completed successfully
+ * @param message
+ *            human-readable description of the outcome
+ * @param timestamp
+ *            when the operation was performed
  */
 public record AliasResult(String aliasName, @Nullable String collections, boolean success, String message,
-                          @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'") Date timestamp) {
+		@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'") Date timestamp) {
 }
--- a/src/main/java/org/apache/solr/mcp/server/collection/AliasResult.java
+++ b/src/main/java/org/apache/solr/mcp/server/collection/AliasResult.java
@@ -16,7 +16,10 @@
  */
 package org.apache.solr.mcp.server.collection;
 
+import com.fasterxml.jackson.annotation.JsonFormat;
+
 import java.util.Date;
+
 import org.jspecify.annotations.Nullable;
 
 /**
@@ -26,17 +29,12 @@ import org.jspecify.annotations.Nullable;
  * Returned by {@link AliasService} methods to communicate the outcome of
  * create, update, and delete operations on Solr aliases.
  *
- * @param aliasName
- *            the alias that was operated on
- * @param collections
- *            the target collection(s) (null for delete operations)
- * @param success
- *            whether the operation completed successfully
- * @param message
- *            human-readable description of the outcome
- * @param timestamp
- *            when the operation was performed
+ * @param aliasName   the alias that was operated on
+ * @param collections the target collection(s) (null for delete operations)
+ * @param success     whether the operation completed successfully
+ * @param message     human-readable description of the outcome
+ * @param timestamp   when the operation was performed
  */
 public record AliasResult(String aliasName, @Nullable String collections, boolean success, String message,
-		Date timestamp) {
+                          @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'") Date timestamp) {
 }
--- a/src/main/java/org/apache/solr/mcp/server/collection/AliasService.java
+++ b/src/main/java/org/apache/solr/mcp/server/collection/AliasService.java
@@ -1,0 +1,207 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.mcp.server.collection;
+
+import io.micrometer.observation.annotation.Observed;
+import java.io.IOException;
+import java.util.Date;
+import java.util.Map;
+import org.apache.solr.client.solrj.SolrClient;
+import org.apache.solr.client.solrj.SolrServerException;
+import org.apache.solr.client.solrj.request.CollectionAdminRequest;
+import org.apache.solr.client.solrj.response.CollectionAdminResponse;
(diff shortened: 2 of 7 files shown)


src/main/java/org/apache/solr/mcp/server/collection/AliasService.java
```diff
@@ -0,0 +1,207 @@
...
+	@McpTool(
+			name = "list-aliases",
+			annotations = @McpTool.McpAnnotations(readOnlyHint = true),
+			description = "List all Solr aliases and the collections they point to")
+	public Map<String, String> listAliases() throws SolrServerException, IOException {
+		CollectionAdminRequest.ListAliases request = new CollectionAdminRequest.ListAliases();
+		CollectionAdminResponse response = request.process(solrClient);
+		// The aliases are returned as a NamedList under the "aliases" key
+		@SuppressWarnings("unchecked")
+		Map<String, String> aliases = (Map<String, String>) response.getResponse().get("aliases");
```
adityamparikh: CollectionAdminResponse has a getAliases() which could be used

src/main/java/org/apache/solr/mcp/server/collection/AliasService.java
```diff
@@ -0,0 +1,207 @@
...
+	 *             if aliasName or collections is blank
+	 * @throws SolrServerException
+	 *             if Solr returns an error
+	 * @throws IOException
+	 *             if there are I/O errors during communication
+	 */
+	@PreAuthorize("isAuthenticated()")
+	@McpTool(
+			name = "create-alias",
+			annotations = @McpTool.McpAnnotations(destructiveHint = false),
```
adityamparikh: the operation is idempotent and it is destructive.

adityamparikh (commented):
Thanks @y-luis-rojo for the contribution. Seems like a useful capability to add. 
Added inline comments as well as overall comments to address: 
1. Looks like the happy path in unit tests need actual assertions.
2. Please add an AliasServiceIntegrationTest that uses Testcontainers to test adding, updating, removing aliases on a new or existing collection. 
3. Please update the docs with this new tool.

src/main/java/org/apache/solr/mcp/server/collection/AliasService.java
y-luis-rojo: Done.

src/main/java/org/apache/solr/mcp/server/collection/AliasService.java
y-luis-rojo: Done.

y-luis-rojo:
> Thanks @y-luis-rojo for the contribution. Seems like a useful capability to add. Added inline comments as well as overall comments to address:
> 
> 1. Looks like the happy path in unit tests need actual assertions.
> 2. Please add an AliasServiceIntegrationTest that uses Testcontainers to test adding, updating, removing aliases on a new or existing collection.
> 3. Please update the docs with this new tool.

Thanks for the comments. I think all are now addressed.

adityamparikh:
An end to end integration tests `AliasServiceIntegrationTest` which uses testcontainers still needs to be added.

y-luis-rojo:
> An end to end integration tests `AliasServiceIntegrationTest` which uses testcontainers still needs to be added.

Sorry, missed to add it. Done.

src/test/java/org/apache/solr/mcp/server/collection/AliasServiceIntegrationTest.java
```diff
@@ -0,0 +1,147 @@
...
+        Map<String, String> aliases = aliasService.listAliases();
+
+        assertThat(aliases).containsEntry(TEST_ALIAS, TEST_COLLECTION);
+    }
+
+    @Test
+    @Order(4)
+    void createAlias_updatesExistingAlias() throws Exception {
+        // Create a second collection to point the alias to
+        String secondCollection = TEST_COLLECTION;
```
adityamparikh: Should this be a separate TEST_COLLECTION_2?

src/main/java/org/apache/solr/mcp/server/collection/AliasService.java
```diff
@@ -0,0 +1,201 @@
...
+            throw new IllegalArgumentException(BLANK_ALIAS_NAME_ERROR);
+        }
+        if (collections == null || collections.isBlank()) {
+            throw new IllegalArgumentException(BLANK_COLLECTIONS_ERROR);
+        }
+
+        CollectionAdminRequest.CreateAlias request = CollectionAdminRequest.createAlias(aliasName, collections);
+        request.process(solrClient);
+
+        return new AliasResult(aliasName, collections, true, "Alias created/updated successfully", new Date());
```
adityamparikh: Should we hardcode success to true?

src/main/java/org/apache/solr/mcp/server/collection/AliasService.java
```diff
@@ -0,0 +1,201 @@
...
+            throws SolrServerException, IOException {
+
+        if (aliasName == null || aliasName.isBlank()) {
+            throw new IllegalArgumentException(BLANK_ALIAS_NAME_ERROR);
+        }
+
+        CollectionAdminRequest.DeleteAlias request = CollectionAdminRequest.deleteAlias(aliasName);
+        request.process(solrClient);
+
+        return new AliasResult(aliasName, null, true, "Alias deleted successfully", new Date());
```
adityamparikh: Should we hardcode success to true?

src/main/java/org/apache/solr/mcp/server/collection/AliasResult.java
```diff
@@ -0,0 +1,42 @@
...
+ *            the target collection(s) (null for delete operations)
+ * @param success
+ *            whether the operation completed successfully
+ * @param message
+ *            human-readable description of the outcome
+ * @param timestamp
+ *            when the operation was performed
+ */
+public record AliasResult(String aliasName, @Nullable String collections, boolean success, String message,
+		Date timestamp) {
```
adityamparikh: Instead of calling it timestamp can we add
@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
for consistency with other Date fields?

adityamparikh (changes requested):

adityamparikh (approved):

adityamparikh:
@epugh Looks good to me. Can you please merge?

epugh:
> @epugh Looks good to me. Can you please merge?

thanks for the review!    Still on vacation, but hoping to get some coding time this week.

epugh:
@y-luis-rojo looks like some lint checks failed...

adityamparikh (approved):

adityamparikh:
I approved. @epugh Could we merge?

adityamparikh:
@epugh re "looks like some lint checks failed" — I dug into this, and **the three red jobs are inherited from the branch point, not caused by this PR.**

The split is the tell: `Build`, `Unit Tests` and `Integration Tests` all pass; only `nativeTest` and the two `native Docker image` jobs fail. This branch points at `582490b`, which predates two native-only fixes that have since landed on `main`:

| Fix | Commit | What it repairs |
|---|---|---|
| #172 | `2809af2` | `DefaultMetaProvider` reflection hint — without it every Spring context refresh fails in native image with "Required no-arg constructor not found" |
| #173 | `9e71cb4` | Docker image tag suffix used by the native integration tests |

`main` measures green natively since those landed, so this is a stale-base problem rather than anything wrong with alias management.

I opened y-luis-rojo/solr-mcp#2 against `feature/alias-management` to merge current `main` in. It is merge-only — no changes to @y-luis-rojo's code. Merging it should turn all three jobs green and make this PR mergeable.

Two things I verified specifically, because a *clean* merge is exactly what hides them:

1. **The `AliasResult` native hint survived the merge.** `main` has been editing that same list (#180 dropped `FieldStats`), so the registration sits between upstream changes on both sides. It is still registered — had it been dropped, the alias tools would have serialized to empty JSON in native image only, which is about the worst failure mode to debug.
2. **`main` gained `SolrNativeHintsTest`**, a JVM-path test pinning those registrations. It passes against this branch.

Locally on JDK 25 the merged result is clean for `compileJava`, `compileTestJava` and `spotlessCheck`.

My approval from earlier still stands on the code itself — this is purely about getting CI honest.

adityamparikh:
Reviewed this as part of a sweep over the open PRs. The alias tools look well-built — blank-input
validation, `AliasResult` registered in `SolrNativeHints`, a Testcontainers integration test, and
MCP hints that match what each tool does (`readOnlyHint` on list, `destructiveHint` on delete).
Those are easy things to miss and they're all here.

**First: the three red native checks are almost certainly not your code.** I fetched the branch to
check rather than guess — it forked from `582490b` (2026-06-19) and contains neither `2809af2` nor
`9e71cb4`, the two native-image fixes that landed on `main` on 2026-08-18. Native was broken on
`main` for everyone before those. A rebase onto current `main` should clear them, and is worth
doing before anyone reads those failures as a problem with alias management.

Two suggestions, both small:

**`AliasResult` uses `java.util.Date`.** The other tool responses use `Instant` —
`CollectionCreationResult` and `SolrHealthStatus` both do. Switching would also let you drop the
`@JsonFormat` pattern, since `Instant` already serialises as ISO-8601.

**`createAlias` takes collections as a comma-separated `String`.** That maps directly onto
SolrJ's signature, which is presumably why it's shaped that way — but the caller here is an LLM,
and a `List<String>` removes the "does whitespace after the comma matter?" question entirely.
Joining with `,` before handing off to `CollectionAdminRequest.createAlias` keeps the SolrJ call
unchanged. This also matches the direction #164 is taking with typed sort clauses.

Neither is blocking. The rebase is the thing worth doing first.

y-luis-rojo:
Sorry the delay. Rebase is done.

adityamparikh:
Re-reviewed the current head (`ad5b94a`). Every earlier inline request is addressed: `getAliases()`, `idempotentHint` on create, `destructiveHint` on delete, the `@JsonFormat` timestamp, `TEST_COLLECTION_2`, and the Testcontainers test. Two things stand between this and merge:

1. **Rebase conflict, one line.** `main` renamed `MCP_RESPONSE_RECORDS` to `MCP_TOOL_RECORDS` in `SolrNativeHints` (and added `SortClause` to it). Add `AliasResult.class` to `MCP_TOOL_RECORDS` when you rebase; that is the only conflict `git merge-tree` reports.
2. **Registration test coverage.** `McpToolRegistrationTest` enumerates Search/Indexing/Collection/Schema for the tool-name uniqueness and annotation checks; the three alias tools are not in it. Please add `addToolNames(AliasService.class, toolNames)` and an `AliasService` annotations case so the invariant covers them.

Two optional nits: `collections` is passed to Solr verbatim, so `"a, b"` reaches the alias API with the space (strip or split/rejoin); and the `getStatus() == 0` branches at lines 166/210 are unreachable because `process()` throws on failure.

The red "CI / Build" is not this PR. The 2026-09-11 re-run died downloading the Gradle wrapper distribution (`Connection reset by peer`) before compiling anything; unit and integration jobs were skipped. A re-run should go green. @epugh, could you kick it off after the rebase?

adityamparikh:
Rebased this branch onto current `main` (`upstream/main` @ `93ffdb7`) to check the conflict. There's exactly one, in `src/main/java/org/apache/solr/mcp/server/config/SolrNativeHints.java` — `main`'s #164 renamed `MCP_RESPONSE_RECORDS` to `MCP_TOOL_RECORDS` (and reworded the comment) after this branch forked off, so the two sides disagree only on that name.

Resolution: keep `main`'s `MCP_TOOL_RECORDS` name/comment, and add `"org.apache.solr.mcp.server.collection.AliasResult"` as the first entry (matching what this branch already registers under the old name).

```diff
-	private static final List<String> MCP_TOOL_RECORDS = List.of(
+	private static final List<String> MCP_TOOL_RECORDS = List.of(
+			"org.apache.solr.mcp.server.collection.AliasResult",
 			"org.apache.solr.mcp.server.collection.CollectionCreationResult",
 			"org.apache.solr.mcp.server.collection.SolrHealthStatus",
 			"org.apache.solr.mcp.server.collection.SolrMetrics", "org.apache.solr.mcp.server.collection.IndexStats",
 			"org.apache.solr.mcp.server.collection.QueryStats", "org.apache.solr.mcp.server.collection.CacheStats",
 			"org.apache.solr.mcp.server.collection.CacheInfo", "org.apache.solr.mcp.server.collection.HandlerStats",
 			"org.apache.solr.mcp.server.collection.HandlerInfo", "org.apache.solr.mcp.server.search.SearchResponse",
 			"org.apache.solr.mcp.server.search.SortClause", "org.apache.solr.mcp.server.schema.SchemaUpdateResult");
```

(Take `main`'s copy of the file wholesale, then just prepend the `AliasResult` line — nothing else in the file conflicts; `README.md` merges cleanly on its own.)

After resolving this and running `./gradlew spotlessApply`, `./gradlew build` passes for me locally — `AliasServiceTest`, `AliasServiceIntegrationTest`, and `SolrNativeHintsTest` all green. (I did hit an unrelated one-off Testcontainers/Docker timeout on the first full run that took out a batch of `http`-profile tests; re-running just those classes passed, so it wasn't caused by this change.)

Steps I used, if useful:
```bash
git fetch upstream main
git rebase upstream/main   # or: git merge upstream/main
# resolve SolrNativeHints.java as above
git add src/main/java/org/apache/solr/mcp/server/config/SolrNativeHints.java
git rebase --continue      # or: git commit
./gradlew spotlessApply
./gradlew build
```

— resolved and verified with Claude Code assistance.

adityamparikh:
Opened #214 with this branch rebased onto current `main` (just the one conflict from #164's rename, resolved as described above) — 438/438 tests passing. All the alias-management work is yours, @y-luis-rojo; this is just carrying it forward so it can merge. Happy to close #214 in favor of this PR once it's rebased, whichever's easier for you and @epugh.

PR #164, opened 2026-07-16
feat(search): replace sort clause maps with a typed SortClause record

adityamparikh:
## Motivation

The `search` tool's `sortClauses` parameter is `List<Map<String, String>>` with magic `"item"`/`"order"` keys. A JSON schema cannot express required keys of a map, so the tool schema MCP clients receive says nothing about what to send — LLM clients guess (`field`/`direction` is a common wrong guess) and sorting silently doesn't apply (`map.get(...)` returns null).

A typed record turns the keys into named schema properties: the contract becomes machine-checkable instead of guessable.

## Changes

- New `SortClause(field, order)` record in the `search` package; `sortClauses` is now `List<SortClause>`.
- LLM-friendly normalization at conversion time: `order` is case-insensitive (`DESC` works) and defaults to `asc` when omitted; an unrecognized order or missing field throws `IllegalArgumentException` naming the offending value, so the client can self-correct on the next call.
- Removed the now-unused `SORT_ITEM`/`SORT_ORDER` constants.
- Registered the record in `SolrNativeHints` (tool parameter records are deserialized reflectively by Jackson, same as response records; list renamed `MCP_RESPONSE_RECORDS` → `MCP_TOOL_RECORDS`).
- Tests: new `SortClauseTest` covering conversion, case normalization, default order, and both rejection paths; existing unit + integration tests updated to the record.

Note: this changes the tool's parameter schema. Map-shaped calls from existing clients would stop matching — but they only "worked" if the client guessed both magic keys correctly, which the schema never told them.

`./gradlew build` passes (unit + Testcontainers integration tests).

### DIFF
--- a/src/main/java/org/apache/solr/mcp/server/config/SolrNativeHints.java
+++ b/src/main/java/org/apache/solr/mcp/server/config/SolrNativeHints.java
@@ -60,18 +60,19 @@ public class SolrNativeHints {
 	}
 
 	/**
-	 * Package-private record types returned by {@code @McpTool} methods. Jackson
-	 * needs reflection access to serialize these as MCP tool responses in native
-	 * image.
+	 * Record types used by {@code @McpTool} methods as responses or parameters.
+	 * Jackson needs reflection access to serialize responses and to bind request
+	 * parameters in native image. Registered by name because some are
+	 * package-private.
 	 */
-	private static final List<String> MCP_RESPONSE_RECORDS = List.of(
+	private static final List<String> MCP_TOOL_RECORDS = List.of(
 			"org.apache.solr.mcp.server.collection.CollectionCreationResult",
 			"org.apache.solr.mcp.server.collection.SolrHealthStatus",
 			"org.apache.solr.mcp.server.collection.SolrMetrics", "org.apache.solr.mcp.server.collection.IndexStats",
 			"org.apache.solr.mcp.server.collection.QueryStats", "org.apache.solr.mcp.server.collection.CacheStats",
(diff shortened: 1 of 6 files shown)


adityamparikh:
Rebased onto `main` (`a84033b`). This one needed real conflict resolution plus a compile fix, so flagging what changed beyond history:

**1. `SolrQuery` moved package (this branch did not compile without it).** SolrJ 10 relocated `SolrQuery` from `org.apache.solr.client.solrj` to `org.apache.solr.client.solrj.request`. This branch was written against SolrJ 9, so `SortClause.java` and `SortClauseTest.java` both had the old import and failed `compileJava` outright after the rebase. Fixed in both.

**2. Reconciled with #176's map validation.** `main` added `SearchService.toSortClause(Map)` plus `SORT_ITEM`/`SORT_ORDER` constants, which is the same problem this PR solves with a type. I removed the map path (constants + private helper) and kept the record, since carrying both would leave two validators for one parameter. Everything else #166 added to `SearchService` — the remediation-hint constants, the `@Nullable` parameters, the eDisMax local-params text from #161 — is preserved.

**3. Updated two test call sites that did not exist when this PR was written.** `SearchServiceIntegrationTest` and `SearchServiceTest` gained `SearchService.SORT_ITEM`/`SORT_ORDER` usages via #166/#176. Git merged them cleanly because they are on different lines; only javac caught them. Both now construct `new SortClause(field, order)`.

**4. `SolrNativeHints`:** kept `main`'s removal of `FieldStats` (#180) and added `SortClause` — tool *parameter* records are Jackson-bound reflectively, so they need the same treatment as response records.

PR #166, opened 2026-07-16
feat(search): append remediation hints to common Solr query errors

adityamparikh:
## Motivation

MCP clients receive a tool's exception message verbatim — the error message is effectively part of the tool's API. Today a raw Solr error like `undefined field foo` gives an LLM client nothing actionable, so it retries blind or gives up. An error that names the follow-up tool arrives at exactly the moment of failure, with perfect targeting, and costs zero tokens until something actually fails.

## Changes

`search` wraps `SolrException` from the query and appends a hint for three recognizable failures:

- **undefined field** → call `get-schema`; mentions dynamic-suffix naming (`name_s`, `price_d`). Solr words this three different ways depending on where the field appears: `undefined field x` (`q`/`fq`), `undefined field: "x"` (facet), and `sort param field can't be found: x` (sort).
- **query syntax error** → `q` uses Lucene syntax; quote/escape special characters, check `{!...}` local params
- **collection not found** → call `list-collections`
- anything unrecognized propagates **unchanged**

`CollectionService`'s collection-not-found error appends the same `list-collections` hint (message prefix preserved, so existing assertions still hold).

Solr's match tokens and the hint text are extracted to named constants, matching the convention in `CollectionService` / `PromptNames`.

### Collection-not-found is matched on the status code, not the message

The first revision of this PR matched the message text for `collection not found` / `can not find` / `404`. **That branch never fired against a real server.** An unknown collection is a 404 whose body is Solr's HTML "not found" page, so SolrJ surfaces it as:

```
Expected mime type in [application/json, text/plain] but got text/html. <p>
  Searching for Solr?<br/>
  You must type the correct path.<br/>
  Solr will respond.
</p>
```

— text containing none of the three tokens. The unit test passed only because it asserted against a fixture string this PR invented.

It now matches `SolrException.ErrorCode.NOT_FOUND.code`. The status code is the only signal that survives — `getMetadata()` and `getRootThrowable()` are both `null` for this response, since the body isn't a Solr JSON error — and using SolrJ's own constant also removes the false-positive risk of matching the bare substring `404` anywhere in a message.

## Tests

**Unit** (mocked `SolrClient`): undefined field, undefined sort field, syntax error, not-found status, and an unrelated Solr error propagating without a hint.

**Integration** (Testcontainers, real Solr) — one per branch: undefined field in `q`, `fq`, facet, and sort; unparseable query; missing collection; and negative `rows` asserting no hint is added.

The integration tests aren't redundant with the unit tests — they're the only ones that can fail. These hints classify Solr's error *text*, which is generated by **solr-core**; only `solr-solrj` is on the classpath. Nothing couples the matchers to Solr's wording at compile time, so a mock-based test can only prove that a stubbed message produces a hint, never that Solr still emits such a message — it would keep passing if a future Solr reworded anything. Each integration test therefore asserts on the *hint* constant, never on the token the matcher uses, so no assertion can pass by tautology. Writing them is what surfaced the dead 404 branch above.

Where a structured signal exists it is preferred over text: 404 → `code()`. Undefined-field and bad-param failures are structurally identical (both a 400 carrying a generic `SolrException`), so text matching there is irreducible — hence the tests.

`./gradlew build` passes (unit + Testcontainers integration tests).

### DIFF
--- a/src/main/java/org/apache/solr/mcp/server/collection/CollectionService.java
+++ b/src/main/java/org/apache/solr/mcp/server/collection/CollectionService.java
@@ -513,7 +513,8 @@ public class CollectionService {
 
 		// Validate collection exists
 		if (!validateCollectionExists(actualCollection)) {
-			throw new IllegalArgumentException(COLLECTION_NOT_FOUND_ERROR + actualCollection);
+			throw new IllegalArgumentException(COLLECTION_NOT_FOUND_ERROR + actualCollection
+					+ ". Hint: call list-collections to see available collections.");
 		}
 
 		// Index statistics using Luke
--- a/src/main/java/org/apache/solr/mcp/server/search/SearchService.java
+++ b/src/main/java/org/apache/solr/mcp/server/search/SearchService.java
@@ -114,6 +114,38 @@ public class SearchService {
 	 * map.
 	 */
 	public static final String SORT_ORDER = "order";
+
+	/**
+	 * Fragments of Solr's own error text that identify a failure we can advise on.
+	 *
+	 * <p>
+	 * These are matched rather than imported because they are produced by
+	 * <em>solr-core</em>, which is not on this server's classpath — only
+	 * {@code solr-solrj} is. Nothing couples them to Solr at compile time, so
+	 * {@code SearchServiceIntegrationTest} pins each one against a real Solr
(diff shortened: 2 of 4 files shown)


epugh:
Yeah, the 404 message may be changing in Solr ;-(

PR #170, opened 2026-07-30
docs(security): document single-tenant deployment model (#169)

### LINKED ISSUE #169: Document that we do not support multi tenancy
In the threat-model work (#160) we discussed whether to explicitly document, at least in v1.0.0 that the mcp does **not** support multi tenancy. I.e. set the expectation that the only intended deployment option is that each user deploys their own mcp server with their own solr credentials, meaning the user understands that all and any action that is allowed by that user in solr-backend (admin permissions etc) will also be available to the mcp. 

By formally documenting this (as already drafted as undecided in the threat model in https://github.com/apache/solr-mcp/pull/160/commits/7ec9c3494b9be2b167f521f7d6b4a4f90df8943c), we will avoid a ton of claimed vulnerability reports claiming that the mcp server allows too much or is insecure in multi tenancy setups.

In a later version, we can consider supporting multi tenancy.

janhoy:
Resolves #169.

Formally documents that the Solr MCP server is **single-tenant** and does not support multi-tenancy: the intended deployment is one instance per user, configured with that user's own Solr credentials, and every action the configured Solr identity is permitted is reachable through the MCP tools by design. Authorization is delegated to Solr, not re-implemented in the MCP layer. Setting this expectation up front avoids "too permissive / insecure for multi-tenant use" reports for behavior that is intentional under this model.

## Changes

- **New `docs/security/deployment-model.md`** — the single authoritative statement of the tenancy model. Concise: states the limitation, that Solr owns authorization, that HTTP-mode OAuth2 is an access gate rather than a tenant boundary, and that sharing one instance across users is unsupported (operator owns the isolation risk). Multi-tenancy noted as possible future work.
- **Pointer links only** (no re-declaration) from the README security list and the STDIO/HTTP transport security docs.

`THREAT_MODEL` is intentionally left untouched — it is handled separately in #160.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ Running in **HTTP mode** — OAuth2, CORS, and the `HTTP_SECURITY_ENABLED` toggl
 **Using it**
 - [Quick start](docs/site/content/pages/mcp/quick-start.md) · [Client setup](docs/site/content/pages/mcp/clients/) — Claude Desktop, Claude Code, VS Code, Cursor, JetBrains, MCP Inspector
 - [Observability](docs/site/content/pages/mcp/observability.md) — OpenTelemetry traces, metrics, logs
-- Security: [STDIO model](docs/security/stdio.md) · [HTTP model](docs/security/http.md) · OAuth2 setup: [Auth0](docs/security/auth0.md) · [Keycloak](docs/security/keycloak.md)
+- Security: [Deployment model (single-tenant)](docs/security/deployment-model.md) · [STDIO model](docs/security/stdio.md) · [HTTP model](docs/security/http.md) · OAuth2 setup: [Auth0](docs/security/auth0.md) · [Keycloak](docs/security/keycloak.md)
 
 **Developing it**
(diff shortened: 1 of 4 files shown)


epugh:
ping @adityamparikh

epugh (approved):
LGTM.   I asked Adi to weigh in as well as he is a core driver of this effort.   We can reevaluate this post 1.0 as a new feature ;-).

adityamparikh (approved):
Agreed — single-tenant is the right stance for 1.0 for both modes:

- STDIO: an instance is spawned per user per client application anyway, so the 1:1 model is inherent to the transport.
- HTTP: secure by default, and the OAuth2 flow authenticates callers — but it's only an access gate; every authenticated caller still acts as the single Solr identity configured at startup.

In both modes the server holds one outbound Solr credential (optionally Basic Auth via `SOLR_USERNAME`/`SOLR_PASSWORD`), so what the tools may do is scoped in Solr, not in the MCP layer at the moment.

+1 to revisiting multi-tenancy (per-request credentials / token exchange) post-1.0.

PR #175, opened 2026-08-02
fix(config): keep empty facet arrays as NamedList, not List

adityamparikh:
`JsonResponseParser` classifies JSON arrays purely by **shape**: an array of
alternating `[String, non-String, ...]` pairs becomes a `NamedList`, anything
else becomes a `List`. An empty array has no shape to inspect, so a facet on a
field that matched zero documents is converted to an empty `List`.

SolrJ's `QueryResponse` casts every `facet_counts/facet_fields` entry to
`NamedList`, so that produces a **`ClassCastException` whenever a faceted field
has no matches** — a plausible query, not an edge case:

```
{"facet_counts":{"facet_fields":{"genre":[],"author":["asimov",3]}}}
                                          ^^ empty -> List -> CCE
```

### Fix

Give the traversal positional context instead of guessing: thread the node path
through `toNamedList`/`convertValue`, and treat anything directly under
`facet_counts/facet_fields` as a `NamedList` regardless of shape. The existing
shape heuristic still covers the other flat-NamedList sites.

### Tests

Adds `JsonResponseParserTest` (5 cases), covering populated facets, the empty
facet, empty/plain arrays *outside* `facet_fields` (which must stay `List`),
and an end-to-end case that feeds the parsed response into a real
`QueryResponse`.

Verified against `main`: reverting only the `JsonResponseParser` change while
keeping the new test fails 2 of the 5 cases, one of them with the
`ClassCastException` above. With the fix, all 5 pass alongside the existing
`JsonResponseParserContentTypesTest`.

`./gradlew spotlessCheck test` is green.

### DIFF
--- a/src/main/java/org/apache/solr/mcp/server/config/JsonResponseParser.java
+++ b/src/main/java/org/apache/solr/mcp/server/config/JsonResponseParser.java
@@ -88,22 +88,33 @@ class JsonResponseParser extends ResponseParser {
 		return List.of(MediaType.APPLICATION_JSON_VALUE, MediaType.TEXT_PLAIN_VALUE);
 	}
 
+	/**
+	 * Path of the object whose direct children are per-field facet arrays. Arrays
+	 * found one level below this path are always NamedLists, regardless of shape.
+	 */
+	private static final String FACET_FIELDS_PATH = "facet_counts/facet_fields";
+
 	@Override
 	public NamedList<Object> processResponse(InputStream body, String encoding) {
 		try {
-			return toNamedList(mapper.readTree(body));
+			return toNamedList(mapper.readTree(body), "");
 		} catch (IOException e) {
 			throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, "Failed to parse Solr JSON response", e);
 		}
 	}
 
-	private SimpleOrderedMap<Object> toNamedList(JsonNode objectNode) {
+	private SimpleOrderedMap<Object> toNamedList(JsonNode objectNode, String path) {
 		SimpleOrderedMap<Object> result = new SimpleOrderedMap<>();
-		objectNode.fields().forEachRemaining(entry -> result.add(entry.getKey(), convertValue(entry.getValue())));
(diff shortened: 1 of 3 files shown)


adityamparikh:
Rebased onto `main` (`a84033b`).

One conflict, in `JsonResponseParser.convertValue`: `main` added a `@Nullable` return annotation (jspecify rollout) on the same signature this PR threads a `path` parameter through. Kept both — `private @Nullable Object convertValue(JsonNode node, String path)`.

No other changes; the fix and its tests are as reviewed. Verified: `./gradlew build` green on JDK 25.

adityamparikh:
Added an end-to-end test and closed a duplicate PR in this one's favour.

**Duplicate:** I opened #185 against the same bug without spotting this PR — my duplicate check
searched issues, not pull requests. #185 is now closed, and issue #182 (which I filed for the
same defect) points here.

Comparing the two implementations, **this one is better** and is what should merge. It anchors on
the full path `facet_counts/facet_fields/<field>`, so a key named `facet_fields` appearing
anywhere else in a response cannot be misread. #185 matched the bare key at any depth. #185 also
covered `facet_queries` and `facet_intervals`, but that turned out to be dead weight — Solr emits
both as JSON objects, which already decode correctly through `convertObject`.

**Ported across (e4d0646):** the one thing #185 had that this PR didn't — an integration test
against real Solr via Testcontainers.

The existing `JsonResponseParserTest` pins decoding at the parser boundary with a hand-written
payload, which is the right place for it. But that leaves one assumption unverified: that a real
Solr genuinely emits `[]` for a zero-hit facet. If Solr ever emitted `{}` instead, those unit
tests would keep passing while the bug they guard had stopped existing in that form. The new test
closes that gap by faceting a filter designed to match nothing and asserting an empty facet map
comes back.

Verified rather than assumed: reverting only the `JsonResponseParser` change on this branch and
re-running makes the new test fail with `java.lang.ClassCastException`. With the fix it passes.
Full build is **378 tests, 0 failures, 7 skipped**.

Issue #183, opened 2026-08-19
create-collection binds every collection to the shared _default configset, leaking schema between collections


adityamparikh:
## Summary

`create-collection` binds each new collection directly to the shared `_default` configset instead
of giving it its own copy. Because `_default` has data-driven (schemaless) mode enabled, the first
collection's **guessed field types are written into the shared managed-schema in ZooKeeper**, and
every collection created afterwards inherits them.

Since Solr field types cannot be modified once created, the second collection is permanently stuck
with the first one's guesses.

## Reproduction

Solr 9, SolrCloud, empty cluster.

```
1. create-collection        name=shows-auto
2. index-json-documents     collection=shows-auto   (schemaless guessing runs)
3. create-collection        name=shows
4. add-fields               collection=shows        (explicit types)
```

Step 4 fails:

```
error processing commands, errors: [{add-field={name=title, type=text_general, ...},
errorMessages=[Field 'title' already exists.]}, {add-field={name=platform, type=string, ...},
errorMessages=[Field 'platform' already exists.]}, ...]
```

`get-schema` on the brand-new, never-indexed `shows` collection returns step 2's guesses:

```json
{"name": "platform",     "type": "text_general"}
{"name": "imdb_rating",  "type": "pdoubles"}
{"name": "release_year", "type": "plongs"}
```

`CLUSTERSTATUS` confirms the sharing:

```
shows       -> configName: _default
shows-auto  -> configName: _default
```

The leak runs in both directions — create a collection with an explicit schema first, and a later
collection silently inherits *those* types instead, so it is not schemaless at all.

## Root cause

[`CollectionService.java:1146`](https://github.com/apache/solr-mcp/blob/main/src/main/java/org/apache/solr/mcp/server/collection/CollectionService.java#L1146):

```java
CollectionAdminRequest.createCollection(name, effectiveConfigSet, effectiveShards, effectiveRf)
        .process(solrClient);
```

with `effectiveConfigSet` defaulting to `_default`. This attaches the collection to that configset
by reference.

For contrast, `bin/solr create -c foo -n _default` **copies** `_default` into a new configset named
after the collection, which is why the same sequence works from the CLI.

## Suggested fix

Have `create-collection` upload a per-collection copy of the source configset (e.g. via
`ConfigSetAdminRequest.Create` with `baseConfigSetName`) and point the new collection at the copy,
matching `bin/solr create` behaviour. The explicit `configSet` parameter should keep its current
[...]

Issue #182, opened 2026-08-19
search: ClassCastException when a facet returns zero buckets


adityamparikh:
## Summary

Any `search` call with `facetFields` that produces **zero facet buckets** fails with:

```
class java.util.ArrayList cannot be cast to class org.apache.solr.common.util.NamedList
```

This is not specific to a field type or a collection — the same field succeeds or fails purely
depending on whether the query matched anything.

## Reproduction

Solr 9 in SolrCloud mode, a collection `shows` with a single-valued `string` field `platform`
(docValues on), 61 documents indexed.

**A — facet has buckets → works**

```json
{"collection": "shows", "query": "*:*", "facetFields": ["platform"], "rows": 0}
```
```
{"numFound":61,...,"facets":{"platform":{"Netflix":20,"HBO Max":7,...}}}
```

**B — same field, filter matches nothing → throws**

```json
{"collection": "shows", "query": "platform:NoSuchPlatform", "facetFields": ["platform"], "rows": 0}
```
```
class java.util.ArrayList cannot be cast to class org.apache.solr.common.util.NamedList
```

Verified end-to-end over MCP STDIO against a published container image, and independently over
the HTTP transport.

## Root cause

Solr serializes an empty facet field as an empty JSON **array**:

```json
"facet_counts": { "facet_fields": { "platform": [] } }
```

`JsonResponseParser.isFlatNamedList` rejects zero-length arrays
([`JsonResponseParser.java:157`](https://github.com/apache/solr-mcp/blob/main/src/main/java/org/apache/solr/mcp/server/config/JsonResponseParser.java#L157)):

```java
int size = arrayNode.size();
if (size == 0 || size % 2 != 0)
    return false;
```

so `convertArray` falls through to the plain-list branch and returns an `ArrayList`. SolrJ's
`QueryResponse.getFacetFields()` then casts that value to `NamedList` and throws.

The heuristic cannot be fixed by simply allowing `size == 0`: a bare `[]` is genuinely ambiguous
between an empty facet NamedList and an empty plain list, and the parser has no context to tell
them apart.

## Suggested fix

Request `json.nl=map` explicitly rather than relying on Solr's `flat` default. Facet fields then
arrive as JSON objects:

```json
"facet_fields": { "platform": {"Netflix": 20} }     // non-empty
"facet_fields": { "platform": {} }                   // empty
```

Both map unambiguously onto `NamedList` through the existing `convertObject` path, which removes
[...]

PR #184, opened 2026-08-19
docs: add a first-collection tutorial teaching schema design

adityamparikh:
## What

Adds `docs/tutorial.md` — a hands-on walkthrough taking a reader from an empty Solr to a
designed, queryable collection, entirely through natural-language conversation with an
assistant. Linked from the README documentation index.

## Why

The docs cover *how to connect a client* and *what each tool does*, but nothing explains
**why field types decide which questions are answerable**. That gap matters here more than in
a normal Solr context: the assistant will happily index anything schemaless, and the user has
no obvious signal that they've traded away faceting and sorting until a question quietly
fails.

So the tutorial indexes the same 61-document dataset twice — once letting Solr guess, once
choosing types on purpose — and puts the two side by side:

| | schemaless | designed |
|---|---|---|
| Documents indexed | 61 of 61 | 61 of 61 |
| `platform` type | `text_general` | `string` + docValues |
| `imdb_rating` type | `pdoubles` (multi-valued) | `pdouble` |
| *"breakdown by platform"* | fails | `Netflix 20, Amazon Prime Video 20, HBO Max 7, …` |

Same data, same question, same tool. The only difference is that someone decided what the
fields meant.

## Verification

Every factual claim was checked end-to-end against Solr 9 in SolrCloud mode, driven through
the MCP server over STDIO — not written from what the tool signatures imply:

- the guessed types (`text_general`, `pdoubles`, `plongs`) and the array-wrapped documents
- all eight search examples and their result counts (1, 2, 20, 31, 15, status 33/28, top-rated
  Game of Thrones 9.2, combined 39)
- the facet buckets quoted in the text
- `check-health` reporting 61 docs, and `get-collection-stats` exposing the cache hit ratios
  the tutorial invites the reader to ask about

## Known issues, documented rather than hidden

A reader following the steps will hit both of these, so the tutorial names them and links the
issues instead of quietly routing around them:

- **#182** — faceting a query that matches nothing throws
  `ClassCastException: ArrayList cannot be cast to NamedList`. This is exactly what the
  schemaless pass runs into.
- **#183** — `create-collection` binds collections to the shared `_default` configset, so
  schemaless guesses leak into every later collection and `add-fields` then fails with
  `Field 'x' already exists`. This is why the tutorial resets Solr midway, and the reset is
  called out as a workaround rather than presented as normal practice.

If either is fixed, the corresponding section should be trimmed — they're marked clearly
enough to find.

## Notes for review

- Uses the raw `shows.json` from `src/test/resources/` via its GitHub URL. Referencing a test
  resource from user docs is a little irregular; happy to move the dataset somewhere more
  deliberate if preferred.
- Docker examples follow the existing `docs/clients/*.md` convention of a locally built image
  rather than a registry reference, since no official image is published yet.
- `**/*.md` is RAT-excluded and the existing `docs/*.md` carry no license header, so this
  follows suit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,7 @@ Running in **HTTP mode** — OAuth2, CORS, and the `HTTP_SECURITY_ENABLED` toggl
 
 **Using it**
 - [Quick start](#quick-start) · [Client setup](docs/clients/) — Claude Desktop, Claude Code, VS Code, Cursor, JetBrains, MCP Inspector
+- [Tutorial: your first collection](docs/tutorial.md) — index a dataset twice, schemaless then with a designed schema, and see why field types matter
 - [Observability](docs/observability.md) — OpenTelemetry traces, metrics, logs
 - Security: [Deployment model (single-tenant)](docs/security/deployment-model.md) · [STDIO model](docs/security/stdio.md) · [HTTP model](docs/security/http.md) · OAuth2 setup: [Auth0](docs/security/auth0.md) · [Keycloak](docs/security/keycloak.md)
 
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -105,17 +105,26 @@ ask them.
 
 > *"Show me the breakdown of shows-auto by platform."*
 
-This is the most ordinary business question imaginable, and it does not work.
-Depending on your Solr version you will get either an empty result or an error —
-see [Known issues](#known-issues) for why the error is currently unhelpful.
+This is the most ordinary business question imaginable, and the answer comes back
+empty:
 
-To understand the cause, look at what Solr decided:
+```json
+{ "numFound": 61, "documents": [], "facets": { "platform": {} } }
+```
+
+Read that carefully, because it is worse than an error. Sixty-one documents
+matched. The query succeeded. Solr simply has no breakdown to give you, and it
+says so without complaining. Nothing here tells you that the *data* is fine and
+the *field type* is the problem — which is exactly the failure mode that makes
+schemaless deceptive.
+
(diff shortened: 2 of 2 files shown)


adityamparikh:
Updated to assume #185 is merged — **this PR now depends on it.**

With the empty-facet fix in place, Step 2 no longer throws. The schemaless facet returns:

```json
{ "numFound": 61, "documents": [], "facets": { "platform": {} } }
```

Which is a better illustration of the point than the exception was. A stack trace reads as
"something is broken and it isn't your fault." A query that succeeds against 61 documents and
returns an empty breakdown is exactly the trap schemaless sets: nothing tells you the data is
fine and the field type is the problem.

Changes:

- Step 2 rewritten around the empty result rather than the crash
- the `#182` entry dropped from Known issues (`#183` and the `sortClauses` note remain)
- the `platform` row sharpened with a concrete consequence — searching `platform:prime`
  matches all 20 Amazon Prime Video shows, so the category has stopped being one value

Re-verified against a build of the #185 branch, driven over the MCP protocol: the empty facet,
the 20-hit token match, the three guessed types (`text_general` / `pdoubles` / `plongs`), and
the array-wrapped documents.

If #185 is rejected or reworked, say so and I'll restore the previous wording — the earlier
revision documented the exception and is still in this branch's history.

adityamparikh:
Correction to my earlier comment: this PR's dependency is **#175**, not #185. #185 was a duplicate of #175 and has been closed; #175 is the PR that makes Step 2 return an empty facet rather than throwing. Nothing in the tutorial text changes — only the PR number this depends on.

PR #185, opened 2026-08-20
fix(search): decode an empty facet as a NamedList, not a List

### LINKED ISSUE #182: search: ClassCastException when a facet returns zero buckets

adityamparikh:
Closes #182.

## The bug

A faceted search whose query matched nothing failed with
`ClassCastException: ArrayList cannot be cast to NamedList` instead of returning an empty
result. Zero matches is an ordinary search outcome, so this is not an edge case — it fires
on properly-typed fields whenever a filter happens to match no documents.

Same collection, same field, only the filter differs:

```java
// works
search("shows", "*:*", null, List.of("platform"), ...)
  -> facets = {platform={Netflix=20, HBO Max=7, ...}}

// throws
search("shows", "platform:NoSuchPlatform", null, List.of("platform"), ...)
  -> ClassCastException: ArrayList cannot be cast to NamedList
```

## Cause

Solr writes a facet field as a flat array under `json.nl=flat`, and writes an empty one as
`[]`. `JsonResponseParser.isFlatNamedList` rejects zero-length arrays:

```java
if (size == 0 || size % 2 != 0)
    return false;
```

so the value fell through to the plain-list branch, and SolrJ's
`QueryResponse.getFacetFields()` — which casts to `NamedList` — threw.

## Why the one-line version is wrong

Allowing `size == 0` in that heuristic looks like the fix and isn't. A bare `[]` is genuinely
ambiguous:

| Response key | Empty form | Must decode as | Because |
|---|---|---|---|
| `facet_counts.facet_fields.<field>` | `[]` | `NamedList` | `QueryResponse.getFacetFields()` casts to it |
| `collections` | `[]` | `List` | `CollectionService.listCollections()` casts to `List<String>` |

Treating every empty array as a NamedList just moves the `ClassCastException` to
`list-collections` against an empty cluster — a first-run experience, not an obscure path.

## The fix

Shape cannot distinguish them, so use the enclosing key. Arrays directly inside
`facet_fields`, `facet_queries` and `facet_intervals` are flat NamedLists by definition,
empty or not. Every other array keeps the existing heuristic untouched.

`json.nl=map` was considered and rejected: `ResponseParser` exposes no hook for query params
and `HttpJdkSolrClient.Builder` has no default-params method, so it would mean wrapping every
request — and `json.nl` is global, silently collapsing the duplicate keys a `NamedList`
permits. (I'd originally suggested it on the issue before reading the code; noting the
correction here.)

**Known gap, documented in the class javadoc:** `facet_ranges` nests its flat list one level
deeper under a `counts` key, so an empty range facet would still decode as a `List`. The
`search` tool does not expose range faceting, so nothing reaches that path today — flagged
rather than silently left.

## Tests

Written first, each watched failing for the right reason before the fix went in:

- **`JsonResponseParserFacetTest`** (new) — empty facet decodes as `NamedList`; populated
  facet still does; empty non-facet array stays a `List`. That last one is the regression
  guard for the trap above. Before the fix: 1 failed, `expected: <NamedList> but was:
  <java.util.ArrayList>`; the other two passed, confirming they pin existing behaviour.
- **`SearchServiceIntegrationTest.facetingAQueryThatMatchesNothingReturnsEmptyFacets`** —
  end-to-end against real Solr via Testcontainers. Before the fix: `java.lang.ClassCastException`.

Full `./gradlew build`: **376 tests, 0 failures, 7 skipped.**

### DIFF
--- a/src/main/java/org/apache/solr/mcp/server/config/JsonResponseParser.java
+++ b/src/main/java/org/apache/solr/mcp/server/config/JsonResponseParser.java
@@ -23,6 +23,7 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Set;
 import org.apache.solr.client.solrj.response.ResponseParser;
 import org.apache.solr.common.SolrDocument;
 import org.apache.solr.common.SolrDocumentList;
@@ -66,8 +67,35 @@ import org.springframework.http.MediaType;
  * every odd-indexed element is a non-{@link String} value. This reliably
  * distinguishes {@code ["term", 5, "term2", 3]} (facet NamedList) from
  * {@code ["col1", "col2"]} (plain string list).
+ *
+ * <p>
+ * <strong>Why shape alone is not enough:</strong> an <em>empty</em> array is
+ * ambiguous. Solr writes a facet with no buckets as {@code []}, and also writes
+ * an ordinary empty list — {@code "collections": []} from the Collections API —
+ * as {@code []}. The first must decode to a {@link NamedList} because SolrJ's
+ * {@code QueryResponse.getFacetFields()} casts to one; the second must stay a
+ * {@link List} because {@code CollectionAdminResponse} callers cast to that. No
+ * rule based on the array can satisfy both, so the parser uses the enclosing
+ * key instead: arrays directly inside {@code facet_fields},
+ * {@code facet_queries} and {@code facet_intervals} are always NamedLists,
+ * empty or not, and every other array falls back to the shape heuristic above.
+ *
+ * <p>
+ * <strong>Known gap:</strong> {@code facet_ranges} nests its flat list one
+ * level deeper, under a {@code counts} key, so an empty range facet would still
+ * decode as a {@link List}. The {@code search} tool does not expose range
+ * faceting, so nothing currently reaches that path.
  */
 class JsonResponseParser extends ResponseParser {
+
+	/**
+	 * Response keys whose immediate children are always NamedLists of counts,
+	 * regardless of how many buckets came back. Solr writes each child as a flat
+	 * array under {@code json.nl=flat}, and writes an empty one as {@code []} —
+	 * which is shape-identical to an ordinary empty list such as
+	 * {@code "collections": []}. Only the enclosing key distinguishes them.
+	 */
+	private static final Set<String> FACET_CONTAINERS = Set.of("facet_fields", "facet_queries", "facet_intervals");
 
 	private final ObjectMapper mapper;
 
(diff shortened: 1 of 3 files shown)


adityamparikh:
Closing as a duplicate of #175, which fixes the same bug and predates this by sixteen days.

I filed issue #182 and opened this PR without finding #175 — my duplicate check searched issues
for "facet" and found nothing, but I never searched pull requests. That's the gap.

**#175 is also the better implementation.** It anchors on the full path
`facet_counts/facet_fields/<field>`, so a key named `facet_fields` occurring elsewhere in a
response can't be misinterpreted; this PR matched the bare key at any depth in the tree. The
extra `facet_queries` / `facet_intervals` coverage here adds nothing either — Solr emits both as
JSON objects, which already decode correctly via `convertObject`.

The one thing this PR had that #175 lacked was an end-to-end test against real Solr. That has
been ported over in adityamparikh/solr-mcp@e4d0646 and verified on #175's branch: reverting only
its `JsonResponseParser` change makes the ported test fail with `ClassCastException`.

Nothing here is lost. Review effort should go to #175.

PR #192, opened 2026-08-31
feat(compose): provision Keycloak with the realm the http profile needs

adityamparikh:
## Problem

The `http` profile authenticates against Keycloak, but nothing started one. A developer had to run the container by hand and then script a realm, a client and an audience mapper before the server would boot. The mapper is easy to miss and fails quietly: Keycloak does not honour the RFC 8707 `resource=` parameter, so without it a token is issued normally and `validateAudienceClaim(true)` answers 401.

## Change

`compose.yaml` gains a `keycloak` service that imports `keycloak/solr-mcp-realm.json` at startup, so the realm, both clients and the mapper exist before the server asks for a token:

| Client / user | Purpose |
|---|---|
| `solr-mcp-service` | Confidential, service accounts enabled: machine-to-machine callers |
| `solr-mcp-client` | Public, redirect URIs for MCP Inspector |
| `testuser` / `testpassword` | The password-grant examples in `keycloak.md` |

Both clients carry the audience mapper for `http://localhost:8080/mcp`. The credentials are development credentials, committed on purpose; a real deployment provisions its own.

The service sits behind the `http` compose profile, which `application-http.properties` activates through `spring.docker.compose.profiles.active`, so a plain `docker compose up -d` for STDIO users does not pull Keycloak or wait on its healthcheck. Start it by hand with `docker compose --profile http up -d`.

**The healthcheck is load-bearing.** The server resolves the issuer while building its JWT decoder and fails to boot if the realm is not answering, so Spring Boot's compose support must wait for the container to be healthy. Keycloak's image ships neither `curl` nor `wget`, so the probe uses bash's `/dev/tcp` against the management port.

`docs/security/keycloak.md` Quick Start points at the imported realm and tells compose users to reuse `solr-mcp-service` instead of creating a client by hand. The manual walkthrough below it binds its own container to the same port, 8180, so anyone following that path should start the server with `SPRING_DOCKER_COMPOSE_ENABLED=false` rather than bringing up a second Keycloak.

## Verification

Against the running container: the realm's OpenID configuration resolves, a `client_credentials` token on `solr-mcp-service` and a password grant on `solr-mcp-client` both carry `aud: http://localhost:8080/mcp`, and `./gradlew build` passes. Not covered: an end-to-end call against a running MCP server.


adityamparikh:
Rebased onto `main` @ `b4ffe18`. One conflict in `docs/security/keycloak.md` (#190 rewrote the Quick Start intro): resolved by keeping both — this PR's compose callout first, then main's runnable-block intro. Added one sentence to the callout: the manual walkthrough and the compose service both bind port 8180, so anyone following the manual path should start the server with `SPRING_DOCKER_COMPOSE_ENABLED=false`. Prerequisites now list Keycloak among the compose-started containers. `./gradlew build` in a clean worktree: 403 tests, 0 failures.

compose.yaml
```diff
@@ -57,6 +57,42 @@ services:
...
+    networks: [ search ]
+    environment:
+      KC_BOOTSTRAP_ADMIN_USERNAME: admin
+      KC_BOOTSTRAP_ADMIN_PASSWORD: admin
+      KC_HEALTH_ENABLED: "true"
+    command: [ "start-dev", "--import-realm" ]
+    volumes:
+      - ./keycloak:/opt/keycloak/data/import:ro
+    healthcheck:
+      test: [ "CMD-SHELL", "exec 3<>/dev/tcp/localhost/9000 && echo -e 'GET /health/ready HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n' >&3 && cat <&3 | grep -q '\"status\": \"UP\"'" ]
```
epugh: @adityamparikh can you jsut investigate this?   It may be we don't have curl or wget available in the image?  so we do this very interesting thing?

epugh: something more like https://github.com/querqy/chorus/blob/main/docker-compose.yml#L299C1-L300C70

epugh (changes requested):
Check one thing on the keyclaok htealh check!

PR #196, opened 2026-09-12
perf(indexing): soft-commit instead of hard-commit, keeping documents searchable

adityamparikh:
`indexDocuments` ended every indexing tool call with `solrClient.commit(collection)` — a hard commit, which fsyncs the segments, so each call waited on the storage device. All four indexing tools (`index-json-documents`, `index-csv-documents`, `index-xml-documents`, `index-markdown-documents`) funnel through it, so that one line governed the whole indexing surface.

Solr's own `_default` configset does the opposite: `autoCommit` at `maxTime` 15000 with `openSearcher=false` — a background hard commit purely to truncate the transaction log — and `autoSoftCommit` at 3000 for visibility. Forcing a synchronous fsync per tool call fought that design.

The commit is now `commit(collection, waitFlush=false, waitSearcher=true, softCommit=true)`.

- **Searchability is preserved.** `waitSearcher=true` keeps the guarantee that matters to a tool caller: the documents are searchable the moment the call returns. Verified 30/30 with zero delay.
- **Durability is unchanged.** The transaction log is written on the `add`, before any commit, so documents survive a crash regardless of commit mode. A hard commit governs how much tlog must be replayed on recovery, not whether data is lost; that housekeeping stays with `autoCommit`.

### Measurement

20 interleaved reps against Solr, same endpoint and document, only the commit parameter varying:

| commit | median | p90 |
|--------|--------|-----|
| none | 4.05 ms | — |
| soft | 8.61 ms | 10.65 ms |
| hard | 18.94 ms | 41.32 ms |

2.2x faster and far tighter — the hard commit's p90 is four times its median, which is fsync variance. Over the MCP tools, 61 single-document calls go from 1853 ms to 585 ms.

### Operator note

A custom configset with `autoCommit` disabled should enable it, or the transaction log grows until something else commits. The `_default` configset already does.

### Tests

`IndexingServiceTest.indexDocuments_SoftCommitsSoDocumentsAreSearchableWithoutForcingAnFsync` pins the soft-commit overload and asserts the hard-commit overload is never used; the existing stubs and verifications in that class move to the four-argument overload. `./gradlew build`: 404 tests, 0 failures.


adityamparikh:
Rebased onto #205 and pushed. **This PR now depends on #205 and should merge after it.** The first four commits shown are #205's; GitHub cannot retarget the base to a branch on the fork, so they appear here until #205 lands.

The rebase was necessary rather than cosmetic. #205 introduces `forward()`, the CSV/XML path. Before the rebase this PR could not reach it, so `a caller that splits a large input across several calls does not pay for a hard commit per call` was only true for JSON and markdown; CSV and XML still issued `setAction(COMMIT, waitFlush=true, waitSearcher=true)`. Neither PR is wrong alone — the gap only exists once both are merged, which is exactly why it survived review twice.

Three fixes on top:

**Soft commit for CSV and XML** (`a4fb3d3`). Measured against Solr directly, same endpoint and document, median of 10: hard commit 22.79 ms vs soft 8.26 ms. Across the MCP tools, 61 single-document CSV calls go from 1853 ms to 585 ms (per extra call 30.4 ms to 9.5 ms). JSON and markdown, which this did not touch, stayed flat — that control is what makes the number trustworthy rather than a warm-cache artifact.

**The size threshold was wrong, and unreachable** (`205a32d`). Observed live: Sonnet 5 in Claude Desktop was asked to index an 86 KB markdown file. After three minutes nothing had reached the server — the model was still emitting the payload as tool arguments, about 76,000 output tokens, past the output budget, so the call could only ever truncate and fail. The instructions said to split inputs larger than `a few hundred KB`; for record data one byte costs roughly one output token, so that threshold is multiples past any output budget. Now ~20 KB per call, with the byte-per-token rule stated so a caller can size its own input, and an instruction to count and batch *before* writing the first call. The guidance also moved into the tool descriptions: `instructions` is returned once at initialize and not every client puts it in the model's context, whereas tool descriptions always are. That reverses `6f15281` deliberately — happy to drop it if you would rather keep the single source.

**Input format vs conversion target** (`4b0815f`). `Send data in the format you already have; when you must convert, emit CSV, and never convert to XML` put those two clauses side by side, so a model holding XML could read `never XML` as applying to its input — and converting XML is the worst possible move, since #205 forwards it to Solr unparsed. Now stated separately, with conversion scoped to formats that have no tool of their own. Only `index-markdown-documents` previously carried an as-is rule; all four do now.

Build green: 392 tests, 0 failures, 7 skipped (the pre-existing `@Disabled` OTLP class).

adityamparikh:
Reduced to one commit. All payload-splitting and format guidance is reverted — the instructions, the tool descriptions and the `index-data` prompt are untouched by this PR. What remains is the commit-mode change, verified against the condition that documents stay searchable.

**Why the guidance went away.** It did not work, and the MCP maintainers consider the shape of it an anti-pattern. Tested live: Sonnet 5 in Claude Desktop, an 86 KB markdown paste. With the guidance in place the model did batch correctly — 15 calls of 10 documents, no truncation — and still took over 15 minutes, because splitting does not reduce total output tokens and each extra call re-reads the whole conversation. Meanwhile [discussion #1197](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/1197) records the consensus as *keep the data plane out of JSON-RPC arguments entirely*, naming "passing entire file content through LLM context" as an anti-pattern, and the [File Uploads WG charter](https://modelcontextprotocol.io/community/working-groups/file-uploads) names "prose instructions asking for base64 strings or local paths" as the problem it exists to remove. Tuning prose was treating a symptom the spec intends to delete; the real answer is the handle-based path ([SEP-2356](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2356), [SEP-2631](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1306)), which is bigger than this PR.

**Why the commit change stays.** Both paths hard-committed — `indexDocuments` via `solrClient.commit(collection)`, and #205's CSV/XML `forward()` via `setAction(COMMIT, waitFlush=true, waitSearcher=true)` — so every tool call waited on an fsync. Solr's own `_default` configset does the opposite: `autoCommit` at 15 s with `openSearcher=false` for tlog truncation, `autoSoftCommit` at 3 s for visibility. We were fighting the configset.

Measured, 20 interleaved reps, same endpoint and document, only the commit parameter varying:

| mode | median | p90 |
|---|---|---|
| no commit | 4.05 ms | 4.64 ms |
| soft commit | 8.61 ms | 10.65 ms |
| hard commit | 18.94 ms | 41.32 ms |

2.2x faster, and far tighter — the hard commit's p90 is four times its median, which is fsync variance. Over the MCP tools, 61 single-document CSV calls go from 1853 ms to 585 ms.

**Searchability is preserved**, which was the condition: `waitSearcher=true` means documents are visible when the call returns, verified 30/30 with zero delay on both the JSON and CSV paths. **Durability is unchanged** — the transaction log is written on the `add`, before any commit, so documents survive a crash in every commit mode; a hard commit governs how much tlog is replayed on recovery, not whether data is lost. One caveat for the docs: an operator running a custom configset with `autoCommit` disabled should enable it.

Still depends on #205 and should merge after it. Build green: 390 tests, 0 failures.

epugh:
What happens if you do a softCommit, but the softCommit plumbing isn't set up?   Can we confirm that IF you use solrcloud you have soft commit set up....    (or normally do at least!).    You may have found a weakness, what happens if you request a soft commit and we haven't set up a trnasaction log etc???/.   This might actually be an issue at the SolrJ layer?

epugh:
> What happens if you do a softCommit, but the softCommit plumbing isn't set up? Can we confirm that IF you use solrcloud you have soft commit set up.... (or normally do at least!). You may have found a weakness, what happens if you request a soft commit and we haven't set up a trnasaction log etc???/. This might actually be an issue at the SolrJ layer?

actually, turns out according to ref guide that you MUST have tlog set up in solrcloud mode!

epugh (approved):
okay, talked it through.

PR #199, opened 2026-09-12
build(test): pin Testcontainers images in the version catalog

adityamparikh:
Stacked on #198 — its commit is included here; merge that first and this diff shrinks to the pinning commit.

### Summary

Pins the Docker images the Testcontainers-based tests start to exact tags, declared once in `gradle/libs.versions.toml`:

```toml
test-image-solr = "solr:9.9.0-slim"
test-image-lgtm = "grafana/otel-lgtm:0.33.0"
```

### Why

- `solr:9.9-slim` was spelled out three times (twice in `build.gradle.kts`, once in `TestcontainersConfiguration`) and is a moving minor tag.
- The LGTM image was `latest`; a cached `latest` on one machine was 0.28.0 while Docker Hub's was 0.33.0, so the same green run meant different images.

### How the pins reach the tests

- `processTestResources` expands them into `src/test/resources/test-images.properties`; `tasks.withType<Test>` also forwards them as the `solr.test.image` / `lgtm.test.image` system properties.
- `TestImages` prefers a non-blank system property, so `./gradlew test -Dsolr.test.image=solr:9.4-slim` and the CI compatibility matrix keep working, and falls back to the expanded resource otherwise — which keeps `@Container static` initialisers working where properties are not forwarded, e.g. `processTestAot`.
- The resource is added to the native test binary with `-H:IncludeResources`.
- `TestImagesTest` fails the build if either pin is missing, floats on `latest`, or is not an exact release tag. It reads the expanded resource rather than system properties, so it cannot interfere with a `-D` override in the same test JVM.
- `AGENTS.md` and `dev-docs/DEVELOPMENT.md` point at the catalog instead of quoting a tag.

### Bumping an image

Edit the two catalog entries. Dependabot's `gradle` ecosystem cannot resolve Docker tags in the catalog, so these stay manual.

`./gradlew build` on Java 25: 405 tests, 0 failures, 0 skipped.


epugh (changes requested):
I'd love to back out the `TestImages.java`, it gives a lot of indirection for not a lot of value.  And I honestly like seeing hte name of the docker image in hte call to TestContainers so I know exactly what solr or lgtm or whatnot is being loaded...  That is the current pattern in `apache/solr`...

adityamparikh:
Backed out `TestImages.java`/`TestImagesTest.java` and the `test-images.properties` resource in a108e8d — `TestcontainersConfiguration` and `OtlpExportIntegrationTest` now call `System.getProperty("solr.test.image"/"lgtm.test.image", "<pin>")` directly at the Testcontainers construction site, same pattern apache/solr uses, so the image name is visible right there. `-Dsolr.test.image` still drives the CI compatibility matrix. The `gradle/libs.versions.toml` pins stay as the declared version (with a comment that the call-site literal needs bumping alongside them) — let me know if you'd rather drop those too.

`./gradlew build`: 403 tests, 0 failures, 0 skipped.

epugh:
thanks!  running build and then will merge.

PR #200, opened 2026-09-13
docs(clients): add a Zed guide and make the HTTP-mode setup self-contained

adityamparikh:
### Summary
Docs only. Three gaps in the client guides, fixed in both `docs/clients/` and the site mirror under `docs/site/content/pages/mcp/clients/`:

- **Zed had no guide.** New `zed.md` covers the settings form and `settings.json` for STDIO (JAR and Docker) and the remote-server entry for HTTP, plus the two things that bite in practice: the schema takes `command`/`args`/`env` with no `source` key, and adding one makes Zed's settings validator reject the entry (verified on Zed 1.16.2); and the 60 s start timeout that a first Docker pull can outrun. Listed in `README.md`, the client index and the site quick start.
- **HTTP Mode sections assumed the reader knew about security.** Each now states that the transport is secured by default and answers 401 until an OAuth2 issuer is configured, and names `HTTP_SECURITY_ENABLED=false` for a local experiment with a pointer to the security docs.
- **Claude Desktop guide did not say why it uses `mcp-remote`.** It now explains that custom connectors are contacted from Anthropic's cloud ([help center](https://support.claude.com/en/articles/11175166-get-started-with-custom-connectors-using-remote-mcp)), so a `localhost` URL is unreachable through the Connectors screen and would need a public HTTPS endpoint with OAuth2 on; the bridge runs locally and needs no tunnel.

### Tests
No code changed; `./gradlew build` and `./gradlew rat spotlessCheck` green. The Zed instructions were exercised against a running server in HTTP mode.


epugh (commented):
So, it's a bit werid that we have docs/clients with user oriented documentation, and then a seperate parralel site/content/pages/mcp/clients with "The official site docs".....    

Chatting with @adityamparikh and he suggested this was the direction that you watned us to go @janhoy to make publishing the future website for solr-mcp easier?   Can you weigh in?   

I'm inclined to merge this PR as is in a few days, since it's a refinement of existing docs + zed docs.   And in another PR, we tackle understanding where the site docs go?

PR #203, opened 2026-09-14
feat(indexing): index many markdown documents in one call

adityamparikh:
### Summary
`index-markdown-documents` took one string and produced exactly one document (#144), so a client with sixty documents needed sixty tool calls, each a full model round trip. That is what made the Markdown path slow: 61 documents took over two minutes end to end while Solr's share was under a second.

The parameter is now `List<String> documents`, one Markdown string per document, each optionally starting with YAML front matter. A whole dataset is one call. Each element goes through the unchanged single-document creator, so front matter, title, headings, body and the content-derived id behave exactly as before.

### Why a signature change and not server-side splitting
There is no standard multi-document Markdown format — Jekyll, Hugo and every parser assume one file per document, and commonmark's front-matter extension only reads a block at the start — so splitting one string into many would be a convention of our own, resting on a heuristic that can misfire on a horizontal rule followed by a `key: value`-looking line. Making the boundary part of the tool schema needs no parsing at all, and the model emits JSON arrays natively. Same approach as #202 for JSON.

### What changed
- `IndexingService.indexMarkdownDocuments(collection, List<String> documents)`; the description says one array element per document and to send many per call.
- `index-data` prompt names the `documents` parameter for the Markdown path.
- `IndexingServiceTest` covers the per-element path; the MCP-level test sends an array. `MarkdownDocumentCreator` is untouched.
- README row.

Independent of #202 and #205; any merge order.

### Tests
`./gradlew build` (Java 25) and `./gradlew nativeTest -Pnative` (GraalVM CE 25.0.2) are green.


adityamparikh:
Superseded by #207, which bundles the `List<String> documents` tool parameter alongside SnakeYAML front matter parsing and multi-record splitting.

PR #201, opened 2026-09-13
test(data): ship the shows sample dataset as CSV, XML and Markdown too

adityamparikh:
### Summary
`shows.json` is the dataset the tutorial and the hackathon card index. The same 61 shows now ship in every format the indexing tools accept, so the walkthrough can be followed in whichever one a reader has:

| File | Shape |
|---|---|
| `src/test/resources/shows.csv` | One row per show. Multi-valued fields (`genres`, `creators`, `cast`, `tags`) are **repeated column headers**; the CSV creator adds one value per non-empty cell under the same field name, so the documents match JSON field for field. |
| `src/test/resources/shows.xml` | Solr update XML: `<add>` with one `<doc>` per record and repeated `<field name="genres">` elements for lists, the format `index-xml-documents` forwards to Solr (#205). |
| `src/test/resources/shows-markdown/<id>.md` | One file per show: every field in YAML front matter, the description as the body under a `# Title` heading. One Markdown document is one Solr document, hence 61 files. |

`ShowsSampleDataTest` pins that the formats the server parses itself (JSON and Markdown) yield the same 61 documents; `ShowsSampleDataIntegrationTest` indexes the JSON, CSV and XML files through the tools into a real Solr and asserts the three collections hold identical documents. CSV and XML go through Solr's own handlers (#205), so this branch carries #205 as a merge until it lands; merge #205 first and this diff shrinks to the data files, the tests and the native include.

The native test binary needs an explicit `-H:IncludeResources` for the three new flavours; Spring AOT only registers `*.json` resources, which is why `shows.json` already worked natively.

### Verification
- `./gradlew build` on Java 25: green, `ShowsSampleDataTest` 4/4.
- `./gradlew nativeTest -Pnative` on GraalVM CE 25.0.2: green including the four new tests (the first run failed on the missing resources, which is what the build change fixes).
- Each format was also pushed through a running server's indexing tool into Solr: 61 of 61 for CSV and XML, one document per Markdown file.
- Rebuilt with #205 merged in: `./gradlew build` green, including the end-to-end equality test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CiUHyyXLTo9ATdgg8eRFZJ

PR #204, opened 2026-09-14
fix(indexing): XML record elements are wrappers, not field prefixes

adityamparikh:
> Stacked on #201 (the CSV/XML/Markdown sample data); the first three commits here are that PR's. Merge #201 first and this diff shrinks to the XML fix.

### Summary
The XML creator flattened every element by its full path from the record element, so

```xml
<shows><show><id>x</id><title>T</title><genres>a</genres><genres>b</genres></show></shows>
```

produced `show_id`, `show_title`, `show_genres` and no `id` at all. Solr minted a UUID per document, re-indexing duplicated instead of replaced, and XML was the one format whose documents did not match the same data in JSON, CSV or Markdown. Anyone following the tutorial with XML had to learn a second set of field names.

The record element (the root for a single document, each repeated child for many) is now a wrapper: its child elements are the fields, named after themselves, nested elements still flatten below that with underscores (`author_name`), and attributes keep the `_attr` suffix: unqualified on the record (`id_attr`), qualified by the element on a child (`name_lang_attr`).

### What changed
- `XmlDocumentCreator.addRecordFields` replaces the prefixed walk for both the single- and multi-document strategies.
- `XmlIndexingTest` expectations drop the record prefix (52 assertions); `ShowsSampleDataTest` now asserts XML equals JSON field for field, the same check CSV and Markdown already passed.
- Tool description and javadoc describe the mapping.

### Compatibility
Field names change for anyone indexing XML today (`book_title` becomes `title`). No test or doc on `main` relied on the old names beyond `XmlIndexingTest` itself, and at 0.0.2-SNAPSHOT the old mapping was a defect rather than a contract.

Part of a set with #202 and #203; independent, any merge order.

### Verification
- `./gradlew build` on Java 25: green.
- `./gradlew nativeTest -Pnative` on GraalVM CE 25.0.2: green.
- Measured against a running server: `shows.xml` indexes 61 documents in one call with `id`, `title`, `platform` and multi-valued `genres`, the same names the JSON path produces.

https://claude.ai/code/session_01CiUHyyXLTo9ATdgg8eRFZJ


adityamparikh:
Closing: this was stacked on #201 and so carried the CSV, XML and Markdown sample files, which are unrelated to the XML fix. The same change, rebased onto `main` as a single commit with only `XmlDocumentCreator`, `XmlIndexingTest` and the tool docs, is the replacement PR linked below. #201's `ShowsSampleDataTest` now expects the fixed mapping and should merge after it.

Replacement: https://github.com/apache/solr-mcp/pull/205

PR #218, opened 2026-09-18
ci: publish nightly Docker image to apache/solr-mcp-nightly

adityamparikh:
## Summary

Follow-up on #15 (nightly build/publish discussion). Re-adds `nightly-build.yml`, deleted in #155, but scoped down to exactly one thing: a Docker Hub preview image (`apache/solr-mcp-nightly`) built from `main`, published daily.

- Fixes the parse-time bug that killed the earlier attempt: `secrets.*` referenced inside an `if:` condition is invalid syntax — GitHub Actions rejects the whole workflow before any job runs (permanent red ✗, `job count: 0`). This version checks secret presence inside a `run:` step instead (same pattern `release-publish.yml` already uses), and fails loudly rather than silently skipping if the Docker Hub secrets are unset.
- Drops the daily GitHub pre-release + source tarball the old version created — that was called out as unwanted noise when #155 removed the workflow.
- Guarded with `if: github.repository == 'apache/solr-mcp'` so scheduled runs on forks (which lack the secrets and shouldn't publish to the `apache/` namespace) no-op instead of failing.
- Reuses the existing `DOCKERHUB_APACHE_USERNAME` / `DOCKERHUB_APACHE_TOKEN` secrets — no new secret required.

**Intentionally out of scope**: publishing source/SBOM snapshots to `nightlies.apache.org`, the other channel raised in #15. That's self-service WebDAV authenticated with an individual committer's own ASF LDAP credentials (`curl -u <asfid> -T file https://nightlies.apache.org/...`, see [nightlies.apache.org/authoring.html](https://nightlies.apache.org/authoring.html)) — not an INFRA-ticket-gated rsync area as the deleted workflow's placeholder assumed. Putting a personal ASF password into a repo secret needs a provisioning decision (role account vs. INFRA-managed credential) before it can be automated; left as a follow-up.

`dev-docs/WORKFLOWS.md` and `dev-docs/DOCKER_PUBLISHING.md` are updated to match, including correcting the guessed `APACHE_NIGHTLIES_USER`/`APACHE_NIGHTLIES_KEY` secret names from the earlier draft (verified those never existed).

## Test plan

- [x] Workflow YAML parses (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [x] Confirmed the only `if:` condition references `github.repository`, not `secrets.*`
- [ ] First scheduled/manual run on `apache/solr-mcp` confirms the image actually pushes (can't test from a fork — no Docker Hub secrets there)


adityamparikh:
One thing worth confirming before/while reviewing this: `apache/solr-mcp-nightly` is a **new** Docker Hub repo name under the `apache` org — unlike `apache/solr-mcp`, which `release-publish.yml` already pushes to successfully with the existing `DOCKERHUB_APACHE_USERNAME`/`DOCKERHUB_APACHE_TOKEN` secrets.

Whether that token can auto-create `apache/solr-mcp-nightly` on first push, or whether it needs to be pre-created by someone with Docker Hub `apache` org admin access, I can't verify from here. The workflow is written to fail loudly rather than silently skip if something's misconfigured, so a first run would surface this clearly either way — but worth a quick check beforehand if convenient, to avoid the first scheduled run just being a known-bad dry run.

@epugh — flagging since this is likely on your side (Docker Hub org access).

adityamparikh:
@epugh — CI is stuck in `action_required` (fork PR, needs a committer to approve the run — I can't self-approve). Also opened the nightly-*jar* counterpart to this in #220, and standardized/reconciled some release docs in #219 and #221 while working through the release checklist for Oct 11.

PR #35, opened 2026-01-07
feat: add automated semantic versioning with git-semver-plugin

adityamparikh:
## Summary
- Add [jmongard/git-semver-plugin](https://github.com/jmongard/Git.SemVersioning.Gradle) for automatic version calculation from conventional commits
- Add `auto-release.yml` workflow for SNAPSHOT builds and changelog accumulation
- Add `cut-release.yml` workflow for manual release creation
- Update `build-and-publish.yml` to use plugin-calculated version
- Comprehensive documentation updates to `WORKFLOWS.md`

### Version Calculation
The plugin analyzes commits since the last tag:
- `fix:` → patch bump (1.0.0 → 1.0.1)
- `feat:` → minor bump (1.0.0 → 1.1.0)
- `feat!:` or `BREAKING CHANGE:` → major bump (1.0.0 → 2.0.0)

### Release Flow
```
PR Merged to main
       │
       ▼
┌──────────────────────┐
│ auto-release.yml     │  SNAPSHOT build
│ Accumulates changelog│  Updates CHANGELOG.md (Unreleased section)
└──────────────────────┘

Manual "Cut Release" (when ready)
       │
       ▼
┌──────────────────────┐
│ cut-release.yml      │  Calculates version from commits
│ Creates v1.0.0 tag   │  Finalizes changelog, creates GitHub Release
└──────────┬───────────┘
           │
           ▼
┌──────────────────────┐
│ build-and-publish    │  Triggered by v* tag
│ Publishes Docker     │  to GHCR
└──────────────────────┘
```

### New Gradle Commands
```bash
./gradlew printVersion      # Show calculated version (e.g., 1.0.0-SNAPSHOT)
./gradlew printChangeLog    # Show auto-generated changelog
```

## Test plan
- [x] `./gradlew printVersion` returns `1.0.0-SNAPSHOT`
- [x] `./gradlew printChangeLog` generates proper changelog from commits
- [x] `./gradlew classes` compiles successfully
- [ ] Verify auto-release workflow accumulates changelog on merge
- [ ] Verify cut-release workflow creates tag and GitHub Release
- [ ] Verify tag creation triggers Docker publishing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #34, opened 2026-01-06
ci: add automated changelog generation with release-please

adityamparikh:
If we use https://www.conventionalcommits.org/en/v1.0.0/, https://github.com/googleapis/release-please will auto generate changelog using github actions https://github.com/googleapis/release-please-action

PR #221, opened 2026-09-18
build(release): wire org.jetbrains.changelog and populate CHANGELOG.md

adityamparikh:
## Summary

Adds the [`org.jetbrains.changelog`](https://github.com/JetBrains/gradle-changelog-plugin) Gradle plugin (2.5.0) and a Keep a Changelog-format `CHANGELOG.md`, so the project has a release-notes source that isn't hand-assembled from git log at release time. Also adds `dev-docs/CHANGELOG_GUIDE.md`, documenting the day-to-day (add an entry to `[Unreleased]` with your PR) and release-time (`patchChangelog -Pversion=X.Y.Z`, `getChangelog`) workflow.

- `[Unreleased]` is seeded **empty** — this is the project's first release, so there's no prior tag to diff against, and hand-picking a few recent commits as if they were a complete backfill would be misleading. `CHANGELOG_GUIDE.md` calls out that the real 1.0.0 notes need a deliberate one-time pass over the entire commit history, done close to the actual release.
- Entries are added by hand to `[Unreleased]` as PRs land. At release time, `patchChangelog -Pversion=<version>` promotes `[Unreleased]` into a versioned section, and `getChangelog --unreleased` / `--project-version=<version>` extracts one version's notes for reuse (e.g. as a GitHub Release body or an announcement email).

## Why manual entries, not commit-driven automation

This deliberately matches the direction already settled on in this repo: PR #34 (release-please) and PR #35 (git-semver-plugin) both proposed fully commit-driven changelog/semver generation and were closed in favor of hand-written entries. `org.jetbrains.changelog` only extracts/patches a manually-maintained section — it does not generate entries from commit messages.

## Note on the `dev-docs/release-process.md` reference

Both the `build.gradle.kts` comment and `CHANGELOG_GUIDE.md` point at the release-process doc as [apache/solr-mcp#157](https://github.com/apache/solr-mcp/pull/157) rather than a file path, since that PR is still open and the file doesn't exist on `main` yet. Once #157 merges this can be simplified to a direct path in a follow-up.

## Testing

- `./gradlew build -x rat` — BUILD SUCCESSFUL, all tests/spotless green. (`rat` fails locally only on pre-existing untracked `.playwright-mcp/*.yml` files unrelated to this change — see `dev-docs/TROUBLESHOOTING.md`-style local-only RAT behavior; doesn't reproduce in CI.)
- `./gradlew getChangelog --unreleased` and `--project-version=1.0.0` — verified extraction against both a populated and the final empty `[Unreleased]` section.
- `./gradlew patchChangelog -Pversion=1.0.0` — ran for real against a scratch copy to confirm the version-cut output, then reverted, since 1.0.0 hasn't actually shipped.


adityamparikh:
@epugh — this one is standalone build-only, doesn't touch the release-process branch or fork. Ready for a look whenever.

